PR #68, opened 2026-06-29
perf(movie): render movie export off the main thread (#58 L-59)

### LINKED ISSUE #58: [Perf] Render/UI hot-path inefficiencies (per-draw depth states, color-picker flood, sync movie export)
**Source:** 2026-06-27 code review (Theme 8). Confirmed. No crash/leak, but real cost.

- [ ] **M-10** `retainInterleavedCPUCopy()` runs unconditionally in `evaluate()` → a full 2nd CPU copy of every VBO, only read on the Metal path → ~2× geometry RAM dead weight on the OpenGL desktop build. `layerGraphics/gl/GLVertexBuffer.cpp:299-303`. Fix: gate on a non-GL renderer being active.
- [ ] **L-37** OIT depth-stencil states allocated per transparent draw call (perf face of #metal-leaks H-3) — `RendererMetal.mm:4088-4092`.
- [ ] **L-48** Sequence panel rebuilds full `str(sequences)` + `get_model` for every object every 500 ms regardless of change. `appkit_sequence_panel.py:297-303`. Fix: gate on a cheap change signal.
- [ ] **L-58** `ColorPicker` fires `set_color`+recolor on every continuous drag tick (no debounce). `ObjectPanel.swift:1402-1406` (and 1442, 1908, 1934). Fix: debounce like `LabeledSlider`.
- [ ] **L-59** Movie export renders each (possibly ray-traced) frame synchronously on the main thread + `usleep` busy-wait → UI beachballs for the whole export. `MovieExportSheet.swift:120-132`.

javierbq:
## L-59 — render movie export off the main thread

`MovieExporter` was `@MainActor`, so each frame's `renderHiResPNG` — a **blocking GPU render that also temporarily reshapes the global core** (`winX/winY` + `PyMOL_Reshape`) — ran on the main thread. Exporting a movie, especially with ray-traced frames (seconds/frame), froze the UI for the whole render.

### Fix: make the export the *exclusive* core user, then render off-main

Because `renderHiResPNG` reshapes global core state, it must not run concurrently with any other core access. So:

1. **`engine.exportRenderActive` flag.** The other two main-thread core accessors gate on it and skip while it's set:
   - the live draw loop (`MetalViewport.draw(in:)`)
   - the feedback poll (`pollFeedback`)

   The export sheet is modal, so no interleaved commands run either — the exporter becomes the **sole** core accessor, and off-main rendering can't race anything.
2. **Dropped `@MainActor` from `MovieExporter`.** Each frame's `cmd.frame` + `renderHiResPNG` + AV/GIF encode (including the old `usleep` wait) now run on a serial `renderQueue` **off the main thread**. Loop state (`idx`/`progress`/`isExporting`) is only touched on the main thread via a hop; the frame index is passed explicitly so the background thread never reads the shared counter.
3. **Flag cleared on completion/cleanup _and_ in `deinit`** — a safety net so dismissing the sheet mid-export can never leave the live viewport frozen.

During export the viewport shows a static frame with a responsive progress bar instead of beachballing; the scene reshape is restored and the live loop resumes when done.

### Verification

- macOS app build succeeds (`** BUILD SUCCEEDED **`).
- ⚠️ **Needs on-device verification** — movie export can't be exercised in the dev harness. On-device, please confirm: export a short **ray-traced** movie and check (a) the UI stays responsive during the render, (b) the output file plays correctly with the right frames, (c) the viewport renders normally afterward, and (d) closing the sheet mid-export doesn't freeze the viewport.

Closes #58 (last remaining item — L-37 fixed in #60, L-58 in #67, M-10/L-48 are GL-/AppKit-only and don't affect the SwiftUI/Metal app).

### DIFF
--- a/swiftui/PyMOLViewer/Panels/MovieExportSheet.swift
+++ b/swiftui/PyMOLViewer/Panels/MovieExportSheet.swift
@@ -21,9 +21,15 @@ import AppKit
 
 // MARK: - Exporter
 
-@MainActor
+// NOT @MainActor: the per-frame core render (renderHiResPNG) + AV/GIF encode run
+// on `renderQueue` so they don't block the UI (#58 L-59). @Published mutations
+// are explicitly hopped back to the main thread; loop control (idx/isExporting)
+// is only touched on the main thread (in renderNext / start / finish).
 final class MovieExporter: ObservableObject {
     enum Format: String, CaseIterable, Identifiable { case mp4 = "MP4", gif = "GIF"; var id: String { rawValue } }
+
+    // Serial: frames render + encode one at a time, off the main thread.
+    private let renderQueue = DispatchQueue(label: "io.raymol.movieexport", qos: .userInitiated)
 
     @Published var isExporting = false
     @Published var progress: Double = 0
@@ -74,6 +80,9 @@ final class MovieExporter: ObservableObject {
         }
 
         engine.pause()           // stop core-driven advance during capture
+        // Claim the core for the exporter: the live draw loop + feedback poll now
+        // skip (they gate on this), so renderHiResPNG can run off-main exclusively.
+        engine.exportRenderActive = true
         isExporting = true
         renderNext()
     }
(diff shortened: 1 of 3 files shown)


javierbq:
## Simulator verification (iPhone 17 Pro, iOS 26.4)

Tested headlessly via the `PYMOL_AUTOEXPORTMOVIE` hook on a 10-model NMR ensemble (`fetch 1d3z; mset 1 -10`).

**The first off-main attempt crashed** — `EXC_BAD_ACCESS` in `_PyObject_Malloc` on the `io.raymol.movieexport` queue. Root cause: `cmd.frame` was driven via `runPython` off-main, but the embedded interpreter uses PyMOL's `PAutoBlock` (single shared thread state, not `PyGILState_Ensure`), so Python from another queue corrupts its heap. Fixed by keeping `cmd.frame` on the main thread and moving only the **Python-free** C++ `renderHiResPNG` + encode off-main (`renderOneOffscreen` verified pure C++/Metal). Pushed as a follow-up commit.

**After the fix — all green:**
- ✅ No crash (app alive through the full export)
- ✅ Valid **10-frame** mp4, 640×360, 0.667 s (10 @ 15 fps)
- ✅ Frames render correctly and are **distinct per NMR state** → per-frame rep rebuild works off-main without the old runloop delay
- ✅ Viewport recovers afterward (live loop resumes, Timeline shows 10/10)

⚠️ Still worth an on-device pass for the **ray-traced responsiveness** win specifically — the simulator has no hardware ray-tracing, so it validates correctness/safety but not the UI-stays-responsive-during-slow-RT-frames behavior that motivated the change.

PR #69, opened 2026-06-29
feat(ios): graduated reset menu (Reset view / effects / Clear session)

javierbq:
## Gap

iOS had **no way to reset app state**. The engine's `clearSession()` (reinitialize → wipe scene/selections/camera + reset all settings to defaults, then re-apply the theme) was solid but only reachable from the **macOS File ▸ Clear Session** menu. So a messed-up scene — or a persisted bad state restored from the autosave, like the stuck filmic tone-map this session chased — had no escape on the phone short of reinstalling.

## What this adds

A graduated **`↺` reset menu** in the iOS toolbar, ordered by blast radius:

| Action | Does | Destructive? |
|---|---|---|
| **Reset view** | `cmd "reset"` (recenter camera) | no |
| **Reset effects** | tone-map / exposure / outline / depth-cue → defaults | no |
| **Clear session…** | reinitialize + re-apply theme + **clear the autosave** | yes → confirmation alert |

### Engine
- **`resetEffects()`** — the Effects-group defaults, now a shared method so the Inspector's "Reset effects" button (#67) and this menu use one source of truth.
- **`clearSessionAndAutosave()`** (iOS) — `clearSession()` **plus** delete the autosave, so a force-quit right after the reset can't restore the cleared/bad state on next launch (the autosave otherwise only clears on a later background-with-empty-scene cycle). This is the actual escape hatch.

Only "Clear session" is destructive and it confirms first; the two light actions reuse already-tested commands.

## Verification
- Simulator build succeeds; the `↺` menu renders in the toolbar (between the theme ◐ and panel/share icons), no toolbar crowding.
- Actions reuse `reset` / `resetEffects` / `clearSession` (the same path macOS uses) + a file delete — low risk. Worth a quick on-device tap-through of Clear session (confirm → empty scene → relaunch stays clean).

### DIFF
--- a/swiftui/PyMOLViewer/Panels/ObjectPanel.swift
+++ b/swiftui/PyMOLViewer/Panels/ObjectPanel.swift
@@ -1923,19 +1923,10 @@ private struct SceneCard: View {
         .buttonStyle(.plain)
     }
 
-    // Effects-group defaults, from layer1/SettingInfo.h. The ~500ms scene-state
-    // poll re-syncs the toggles/sliders to these values after the sets land.
-    private func resetEffects() {
-        let defaults: [(String, String)] = [
-            ("metal_outline", "0"),
-            ("metal_outline_width", "1.4"),
-            ("metal_outline_color", "0x000000"),
-            ("metal_tonemap", "0"),
-            ("metal_exposure", "1.0"),
-            ("depth_cue", "1"),
-        ]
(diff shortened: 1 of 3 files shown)


javierbq:
Superseded by https://github.com/javierbq/RayMol/pull/70, which combines this reset menu with the new long-press context menu into one 'iOS context menus' PR.

Issue #123, opened 2026-07-08
iOS SwiftUI app: no documented/scripted path to populate deps_ios/


weitzner:
While setting up a new dev machine (MacPorts, not Homebrew) to build and test both the macOS and iOS SwiftUI apps, the macOS side had a clean, scripted setup path (`scripts/fetch_macos_python.sh` + `swiftui/build_macos.sh`), but the iOS side does not.

`swiftui/PyMOLBridge.xcconfig` and `swiftui/build_ios.sh` / `appkit/CMakeLists_iOS.cmake` expect `deps_ios/` (gitignored) to already contain:

- `deps_ios/Python.xcframework/{ios-arm64,ios-arm64_x86_64-simulator}/Python.framework` — a BeeWare-style CPython 3.13 xcframework (headers + dylib + stdlib)
- `deps_ios/install` and `deps_ios/install_device` — freetype/libpng cross-compiled for iOS simulator and device
- `deps_ios/numpy-ios/{simulator,device}/numpy` — buildable via `scripts/build_numpy_ios.sh`, but that script itself requires `deps_ios/Python.xcframework` to already exist

There's no script anywhere in the repo (`scripts/`, `appkit/`, `.github/workflows/`) that fetches or builds `Python.xcframework` or cross-compiles freetype/libpng into `deps_ios/install{,_device}`. `.github/workflows/build.yml` has no iOS job, so there's no CI recipe to reverse-engineer either.

`docs/superpowers/plans/2026-06-09-pymol-ipad-metal-mvp.md` mentions the BeeWare Python.xcframework and cross-compiled freetype/libpng but doesn't give fetch/build commands, and even calls out the BeeWare stdlib layout as a top risk/time-sink for whoever sets this up next.

**Ask:** a `scripts/fetch_ios_python.sh` (or similar) analogous to `scripts/fetch_macos_python.sh`, plus either a documented cross-compile recipe or vendored prebuilt freetype/libpng slices for `deps_ios/install{,_device}`, so a new contributor can build+run the iOS target without reverse-engineering the original bring-up.

(Found while working on #app-icon and macOS MacPorts-portability fixes — happy to file those as separate PRs.)

javierbq:
[...]

PR #124, opened 2026-07-08
Add Icon Composer app icon and wire it into iOS + macOS targets

weitzner:
## Summary
- Adds the new Icon Composer `.icon` bundle (`swiftui/PyMOLViewer/Resources/RayMol.icon`) as the app icon source for both `PyMOLViewer_iOS` and `PyMOLViewer_macOS`.
- Sets `ASSETCATALOG_COMPILER_APPICON_NAME = RayMol` for both targets' Debug/Release configs.
- Leaves the existing `AppIcon.appiconset` in place as a fallback (unused by either target now, but not deleted).

## Test plan
- [x] Built `PyMOLViewer_macOS` via `xcodebuild` and confirmed `RayMol.icns` is compiled into the app bundle from the new Icon Composer assets (verified visually — pink/orange gradient, flower + "R" glyph).
- [x] Launched the built macOS app and confirmed it runs.
- [ ] iOS target not build-verified in this environment (deps_ios/ isn't set up locally — tracked separately in #123, unrelated to this change).


javierbq (approved):
Reviewed and validated end-to-end. Built both targets with Xcode 26.4:

- **macOS**: `RayMol.icns` compiles from the Icon Composer bundle; verified the new icon (pink→orange gradient, p-orbital petals, white "R") in the Dock of a running instance in a disposable macOS VM.
- **iOS**: asset catalog compiles the `RayMol` renditions; installed on the iPhone 17 Pro simulator and confirmed the correct icon on the home screen. (The iOS build only needs the prebuilt `build_ios/libpymol_core.a` present — orthogonal to this change.)

pbxproj changes are safe: the `PyMOLViewer.app`→`RayMol.app` product-reference rename is cosmetic (`PRODUCT_NAME = RayMol` already set), and `PyMOLBridge.xcconfig` moving into the auto-generated "Recovered References" group is a navigator-only reshuffle — it's still wired as `baseConfigurationReference`. LGTM.

Issue #2, opened 2026-03-25
Feature: Add unit tests for AppKit host and AI chat system


javierbq:
## Description
The AppKit host, Metal renderer, AI chat system, and Python UI modules have no automated tests. This makes development slow and fragile — bugs are only caught through manual testing.

### AI Chat System
- \`ai_chat.py\`: test \`_parse_structured_response()\` with valid JSON, malformed JSON, plain text
- \`ai_tools.py\`: test tool definitions schema, \`execute_tool()\` dispatch
- \`ai_pdb_search.py\`: test \`search_pdb()\` with mock HTTP responses
- \`ai_system_prompt.py\`: test SYSTEM_PROMPT is valid non-empty string

### UI Modules (headless where possible)
- \`appkit_mouse_panel.py\`: test \`_build_mode_array()\`, \`_code_for()\`, color definitions
- \`appkit_object_panel.py\`: test menu construction logic
- \`metal_pick.py\`: test NDC-to-world math with known view matrices

### Metal Renderer
- \`SceneRenderMetalSelections()\`: test with known selection coordinates
- \`ExecutiveGetSelectionCoords()\`: test with loaded molecule

### Integration Tests
- Launch app, load molecule, verify object panel updates
- Click on atom, verify selection created
- Send AI chat message, verify response parsing

## Framework
Use \`pytest\` for Python tests, \`catch2\` for C++ tests (already configured in \`layerCTest/\`).

## Files
- New: \`testing/tests/appkit/\` directory
- New: \`testing/tests/ai_chat/\` directory

Issue #1, opened 2026-03-25
Bug: Ray/Draw button doesn't display rendered image in viewport


javierbq:
## Description
The Draw/Ray dropdown (top-right) has options: Draw, Ray, Ray (High Quality), Draw (Width x Height). When selected, the ray trace runs but the result is not displayed in the Metal viewport.

## Expected
After ray tracing completes, the rendered image should be displayed in the viewport (replacing the real-time view until the user clicks/rotates).

## Root Cause
Ray tracing produces an image via `SceneRay()` → `ScenePNG()` which stores pixels in `CScene::Image`. The OpenGL path displays this via `SceneCopy()` → `glDrawPixels()`. The Metal path has no equivalent image display mechanism.

## Possible Fix
After ray trace, read the `CScene::Image` buffer and blit it as a textured quad through the Metal renderer, or display it in an NSImageView overlay.

## Files
- `layer1/SceneRender.cpp` — `SceneRenderMetal()`
- `layer5/main_appkit.mm` — Metal view

PR #125, opened 2026-07-08
Make macOS native build portable across package managers

weitzner:
## Summary
- `PyMOLBridge.xcconfig` and `appkit/CMakeLists.txt` hardcoded `/opt/homebrew` for freetype/libpng/glm/libomp, which breaks the native macOS SwiftUI build for MacPorts users (`/opt/local`). Introduces an overridable `PYMOL_EXTERNAL_PREFIX` (CMake cache var + xcconfig variable, defaulting to `/opt/homebrew`), with an optional gitignored `swiftui/PyMOLBridge.local.xcconfig` for per-developer overrides and a `PYMOL_EXTERNAL_PREFIX` env var read by `swiftui/build_macos.sh`.
- Fixes a latent macro mismatch: the CMake core build always compiles `libpymol_core.a` with `_PYMOL_NO_OPENGL`/`_PYMOL_METAL_ONLY` for this target, but the Xcode-side compile of `PyMOLBridge.mm` only defined that macro for iOS SDKs, never `macosx`. This let the macOS bridge file silently compile down the real-GL code path (masked on Homebrew machines that happen to have `GL/glew.h`/`GL/glut.h` on the include path) instead of agreeing with the Metal-only core it links against.

## Test plan
- [x] Installed MacPorts equivalents (`glm glew libpng freetype libomp`) on a machine with no Homebrew.
- [x] Populated `deps_macos/python-standalone` via `scripts/fetch_macos_python.sh`.
- [x] Built `libpymol_core.a` via `PYMOL_EXTERNAL_PREFIX=/opt/local swiftui/build_macos.sh`.
- [x] Built `PyMOLViewer_macOS` via `xcodebuild` using a local `PyMOLBridge.local.xcconfig` pointing at `/opt/local` — succeeded end-to-end.
- [x] Launched the built app and confirmed it runs.
- [x] Confirmed the default (no override) behavior is unchanged — `PYMOL_EXTERNAL_PREFIX` defaults to `/opt/homebrew` everywhere.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,9 @@ dd_ios
 # per-user Xcode UI state (should never be tracked)
 xcuserdata
 
+# per-developer package-manager prefix override (see PyMOLBridge.xcconfig)
+PyMOLBridge.local.xcconfig
+
 # scratch test structures / session transcripts at repo root
 /*.cif
 /*.pdb
--- a/appkit/CMakeLists.txt
+++ b/appkit/CMakeLists.txt
@@ -38,6 +38,11 @@ endif()
 # Homebrew Python. Builds only pymol_core (no AppKit executable).
 option(PYMOL_METAL_ONLY "Native macOS Metal-only build (no OpenGL)" OFF)
 
+# External package-manager prefix for build-time headers/libs not vendored in
+# deps_macos/deps_ios (freetype, libpng, glm, GLEW). Homebrew by default;
+# override with -DPYMOL_EXTERNAL_PREFIX=/opt/local for MacPorts.
+set(PYMOL_EXTERNAL_PREFIX "/opt/homebrew" CACHE PATH "External package-manager prefix (Homebrew/MacPorts)")
+
 # Options
 option(PYMOL_LIBXML "Enable libxml2 (COLLADA)" ON)
 option(PYMOL_VMD_PLUGINS "Enable VMD plugins" ON)
@@ -51,7 +56,7 @@ if(PYMOL_METAL_ONLY)
     # PyMOLBridge.xcconfig. Python headers come from the staged
     # python-build-standalone tree (PYMOL_PYTHON_INCLUDE_DIR); png/freetype
     # headers from Homebrew (build-time only).
-    set(BREW_PREFIX "/opt/homebrew")
+    set(BREW_PREFIX "${PYMOL_EXTERNAL_PREFIX}")
     set(Python3_INCLUDE_DIRS "${PYMOL_PYTHON_INCLUDE_DIR}")
     set(Python3_NumPy_INCLUDE_DIRS "")
     set(PNG_INCLUDE_DIRS "${BREW_PREFIX}/include/libpng16" "${BREW_PREFIX}/include")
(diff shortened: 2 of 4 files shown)


javierbq (changes requested):
Thanks for this — the ABI/macro fix in particular is a real catch, and the portability effort is welcome. A few notes from review. Requesting changes only for the two OpenMP gaps below; the macro fix itself is good to go.

## The macro flip is correct and safe to merge ✅

Renaming `PYMOL_IOS_FLAGS` → `PYMOL_NOGL_FLAGS` and adding the `[sdk=macosx*]` variant with `-D_PYMOL_NO_OPENGL -D_PYMOL_METAL_ONLY` is the right fix. Before this, the macOS bridge got neither define and compiled the real-GL branch of `layer0/os_gl.h` while `libpymol_core.a` (built `-DPYMOL_METAL_ONLY=ON`, i.e. `_PYMOL_NO_OPENGL`) compiled the stub branch — a genuine ODR/type mismatch that happened to be benign only because the two `struct GLFramebufferConfig` definitions coincide byte-for-byte on Homebrew toolchains. `PyMOLBridge.mm` references no GL symbols directly and the core types it touches (`PyMOLGlobals`, `RendererMetal`) have no macro-conditional members, so the flip compiles cleanly and brings the bridge into agreement with the core it links against. Since the new macro isn't gated on the prefix, your MacPorts build already exercised the exact path the Homebrew default now takes.

Consider splitting this into its own commit — it's an orthogonal, higher-value fix that stands on its own regardless of the package-manager work.

## The portability half is incomplete for OpenMP — two gaps to close

The build "succeeds and runs" on MacPorts, but OpenMP is silently dropped along the way, so the app ends up running the serial surface-sampling path without any hard error to signal it:

1. **`appkit/CMakeLists.txt:259-260`** — the `omp.h` discovery isn't parameterized:
   ```cmake
   find_path(LIBOMP_INCLUDE omp.h
       HINTS /opt/homebrew/opt/libomp/include /usr/local/opt/libomp/include)
   ```
   This runs for the metal-only core (`PYMOL_OPENMP` defaults ON). The hints list only the two Homebrew prefixes and never consult `PYMOL_EXTERNAL_PREFIX`, so on MacPorts `find_path` fails, the `WARNING` at line 266-267 fires, and `pymol_core` builds serial. Suggest adding `${PYMOL_EXTERNAL_PREFIX}/include/libomp` (and any MacPorts libomp include dir) to the HINTS.

2. **`swiftui/PyMOLBridge.xcconfig`** — the derived libomp default bakes in Homebrew's keg layout:
   ```
   PYMOL_LIBOMP_STATIC = $(PYMOL_EXTERNAL_PREFIX)/opt/libomp/lib/libomp.a
   ```
   The `/opt/libomp/lib` infix is Homebrew-specific; your own example comment correctly shows MacPorts as `/opt/local/lib/libomp/libomp.a`, a shape this derivation can't produce. So setting only `PYMOL_EXTERNAL_PREFIX` yields a nonexistent path — the user *must* also override `PYMOL_LIBOMP_STATIC`. That's a fine escape hatch, but it means the "just set the prefix" model doesn't hold for libomp; worth calling out explicitly in the comment (which you partly do) or deriving from a layout that works for both.

Together these interact: fix #1 lets the core build with OpenMP, and fix #2 lets the link find the runtime.

## Minor / optional

- **`swiftui/build.sh:26`** already had its own `HOMEBREW_PREFIX` override, so the repo now has two divergent env-var names for the same concept. That script targets the abandoned `build_appkit/` path (not `build_macos.sh`) and isn't referenced by any current skill/CI — probably worth a `DEPRECATED` header or removal in a follow-up rather than leaving the mismatch.
- The `/opt/homebrew` default now lives in three files (CMake cache var, xcconfig, `build_macos.sh`). The shell→CMake pair stays in sync via `-D`, but the xcconfig default is read independently by Xcode and can drift — a "keep in sync" comment would help.
- Nice touch that `scripts/bundle_macos_dylibs.py` already handles `/opt/local`, so the bundling stage needed no changes.

Heads-up for other reviewers: none of this is covered by CI — the native macOS SwiftUI/Metal build (`build_macos.sh` → `xcodebuild`) is local-only, so the `build.yml` checks don't touch these files.

javierbq (approved):
LGTM!

Issue #185, opened 2026-07-11
iOS/iPadOS command console: smart quotes/dashes break Python and PyMOL commands with strings


weitzner:
Found while verifying numpy on a physical iPad build: typing a command containing a quote character in the iOS/iPadOS command console (`swiftui/PyMOLViewer/Panels/CommandPanel.swift`) gets the straight `'`/`"` silently replaced by iOS's smart-quotes autocorrection with curly `'`/`'`/`"`/`"`, producing invalid Python/PyMOL syntax. Reported symptom: "the double and single quote characters aren't the plain text quote characters so they throw invalid character errors."

This affects any command with a string literal — e.g. `select foo, resn HIS` is fine (no quotes), but anything going through `/`-prefixed inline Python (`/import numpy; print(numpy.__version__)`) or PyMOL commands taking quoted string arguments breaks immediately.

**Root cause**: the iOS variant of `CommandTextField` (`CommandPanel.swift:271`, a plain SwiftUI `TextField`) already sets:
```swift
.textInputAutocapitalization(.never)
.autocorrectionDisabled(true)
```
but neither of those covers iOS's smart-quotes/smart-dashes/smart-insert-delete substitution — those are separate `UITextInputTraits` properties (`smartQuotesType`, `smartDashesType`, `smartInsertDeleteType`) that plain SwiftUI `TextField` doesn't expose a modifier for. `autocorrectionDisabled` only controls spelling autocorrection, not text substitution.

**Why this isn't a one-line fix**: the code comment directly above (`CommandPanel.swift:265-270`) explains this SwiftUI-native `TextField` implementation replaced an earlier `UIViewRepresentable`-wrapped `UITextField`/`UITextView`, specifically because that version "didn't submit, focused unreliably, and got covered by the keyboard at the bottom of the panel." Properly fixing the smart-quotes issue means going back to a UIKit wrapper (the only way to set `smartQuotesType = .no` etc.) — so any fix needs to carefully preserve submit-on-return, reliable focus, and keyboard-avoidance behavior while adding the missing text-input traits, not just revert the earlier change.

**Repro**: on the iOS/iPadOS command console, type a command with a quote character, e.g.:
```
/import numpy; print("NUMPY_OK", numpy.__version__)
```
The straight quotes get silently converted to curly quotes, and the command fails with a syntax/invalid-character error.

javierbq:
@weitzner you mentioned this bug in #186, is this closed?

weitzner:
It is not

PR #186, opened 2026-07-11
Get the iOS/iPadOS build working, bring it to parity with the macOS build

weitzner:
## Summary

Gets the iOS/iPadOS build working end-to-end for the first time on a MacPorts
machine, and brings its developer experience as close to the macOS build's as
the platforms allow.

- **Parameterize the iOS CMake branch's Homebrew prefix by `PYMOL_EXTERNAL_PREFIX`**
  — `appkit/CMakeLists.txt`'s `PYMOL_IOS` branch hardcoded `/opt/homebrew`,
  the same class of bug already fixed on the macOS side (#125). Verified with
  a real device build: went from `glm/vec3.hpp not found` to a clean link.
- **Two real bugs in `scripts/build_numpy_ios.sh`**, found while verifying the
  above, each independently confirmed with a full cross-compile (not just the
  precheck):
  - The build-tools precheck did `import meson, ninja` — but `pip install meson`
    installs as `mesonbuild`, so the check failed even with meson correctly
    installed. Confirmed the correct import name with a clean venv install
    before fixing.
  - `PYHOST` defaulted to a hardcoded `/opt/homebrew/bin/python3.13`; now
    derives from `PYMOL_EXTERNAL_PREFIX` like everywhere else.
- **`scripts/setup_ios_deps.sh`** — a one-command wrapper running all four
  `deps_ios/` setup scripts in order, giving iOS the same one-command feel as
  `fetch_macos_python.sh` on macOS (it unavoidably has more moving pieces — a
  cross-compiled dependency set vs. a single vendored Python — but the
  *experience* of setting it up is now the same: one command).
- **`swiftui/README.md`** — documents the iOS build end-to-end for the first
  time: the simulator fast-path (no Apple Developer account needed, matching
  macOS's zero-friction local build) vs. physical-device deployment (the full
  first-time signing dance: Developer Mode, Xcode account, Program License
  Agreement, device registration — none of which was written down anywhere
  before this).

One thing this PR does *not* try to paper over: physical-device deployment
genuinely can't match macOS's simplicity, because it's Apple's platform
policy (real iOS hardware requires Developer Program signing), not a gap in
this build system. Documented thoroughly instead of pretended away.

Also filed while doing this work, both out of scope for this PR:
- #185 — iOS command console: smart quotes/dashes silently corrupt any
  command containing a string literal (found while verifying numpy on
  device). Real bug, but fixing it properly means redoing a UIKit text-field
  wrapper carefully enough not to reintroduce a prior regression (see issue
  for detail) — not a quick fix, deserves its own PR.

## Test plan
- [x] `PYMOL_EXTERNAL_PREFIX=/opt/local swiftui/build_ios.sh device` — went
      from a missing-glm build failure to a clean link.
- [x] Verified the `mesonbuild` import fix with a clean `pip install meson
      ninja` into a throwaway venv before changing the script.
- [x] Full re-run of the fixed `build_numpy_ios.sh` (no manual `PYHOST`
      override) produced both `deps_ios/numpy-ios/{simulator,device}/numpy`
      slices.
- [x] `scripts/setup_ios_deps.sh` run end-to-end, all four steps succeeded.
- [x] Built, installed, and launched on the iOS Simulator (iPad Pro 11-inch
      M5) — confirmed rendering (cartoon representation, sequence viewer,
      command console, Theme Studio all functional) via screenshot.
- [x] Built, signed (Outpace Bio Inc team), installed, and launched on a
      physical iPad (iPad17,1) — first on-device run on this Mac.
- [x] Re-verified the physical device build with the corrected numpy/Biopython
      bundle: `numpy.__version__` prints `2.4.6` on-device, matching the
      cross-compiled version.

### DIFF
--- a/appkit/CMakeLists.txt
+++ b/appkit/CMakeLists.txt
@@ -83,7 +83,7 @@ else()
     # iOS: headers-only mode for static library compilation.
     # Use BeeWare Python.xcframework headers for iOS builds;
     # actual linking happens in the Xcode project.
-    set(BREW_PREFIX "/opt/homebrew")
+    set(BREW_PREFIX "${PYMOL_EXTERNAL_PREFIX}")
     set(PYMOL_IOS_DEPS "${CMAKE_SOURCE_DIR}/../deps_ios" CACHE PATH "iOS dependencies root")
 
     # Python headers from BeeWare Python.xcframework (3.13)
--- a/scripts/build_numpy_ios.sh
+++ b/scripts/build_numpy_ios.sh
@@ -25,7 +25,7 @@ set -euo pipefail
 NUMPY_VERSION="${NUMPY_VERSION:-2.4.6}"
 DEPLOY="${IOS_DEPLOYMENT_TARGET:-16.0}"
 ROOT="$(cd "$(dirname "$0")/.." && pwd)"
-PYHOST="${PYHOST:-/opt/homebrew/bin/python3.13}"
+PYHOST="${PYHOST:-${PYMOL_EXTERNAL_PREFIX:-/opt/homebrew}/bin/python3.13}"
 WORK="$(mktemp -d)"
 OUT="$ROOT/deps_ios/numpy-ios"
 
@@ -33,7 +33,7 @@ PY_SIM_HDR="$ROOT/deps_ios/Python.xcframework/ios-arm64_x86_64-simulator/Python.
 PY_DEV_HDR="$ROOT/deps_ios/Python.xcframework/ios-arm64/Python.framework/Headers"
 
 echo ">> host python: $PYHOST"; "$PYHOST" --version
-"$PYHOST" -c "import meson, ninja" 2>/dev/null || {
+"$PYHOST" -c "import mesonbuild, ninja" 2>/dev/null || {
   echo "ERROR: install build tools first:  $PYHOST -m pip install meson ninja cython build"; exit 1; }
 
 echo ">> downloading numpy $NUMPY_VERSION sdist"
--- a/scripts/setup_ios_deps.sh
+++ b/scripts/setup_ios_deps.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+# setup_ios_deps.sh — populate deps_ios/ end-to-end: embedded Python, cross-
+# compiled freetype/libpng, numpy, and Biopython, in dependency order. Mirrors
+# the one-command feel of fetch_macos_python.sh for the iOS side, which
+# unavoidably has more moving pieces (a cross-compiled dependency set, not a
+# single vendored Python). Re-run to refresh everything; each step is
+# individually idempotent.
+#
+# MacPorts (or any non-Homebrew prefix): export PYMOL_EXTERNAL_PREFIX=/opt/local
+# first — see swiftui/README.md's "Package-manager portability" section.
+set -euo pipefail
+
+ROOT="$(cd "$(dirname "$0")/.." && pwd)"
+
+echo "=== 1/4: fetch_ios_python.sh ==="
+"$ROOT/scripts/fetch_ios_python.sh"
+
+echo ""
+echo "=== 2/4: build_ios_deps.sh (freetype + libpng cross-compile) ==="
+"$ROOT/scripts/build_ios_deps.sh"
+
+echo ""
+echo "=== 3/4: build_numpy_ios.sh ==="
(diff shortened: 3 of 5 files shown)


javierbq:
## Code Review: iOS/iPadOS build parity

A build-infrastructure + documentation PR that gets the iOS/iPadOS build working end-to-end on a non-Homebrew (MacPorts) machine and closes the DX gap with the macOS build. **Low-risk, high-value** — I verified each change against the checkout and found no correctness issues.

### Correctness — verified ✓

- **`appkit/CMakeLists.txt:86`** `/opt/homebrew` → `${PYMOL_EXTERNAL_PREFIX}` is safe. The variable has a `CACHE PATH` default of `/opt/homebrew` (line 44), so the manual `cmake` invocations in the file header (which don't pass `-D`) still work unchanged on Homebrew — no regression. It now matches the `PYMOL_METAL_ONLY` branch (line 59), which is exactly the parity claimed (#125-class fix).
- **`build_numpy_ios.sh:36`** `import meson` → `import mesonbuild` is a genuine fix. The meson PyPI package exposes `meson` only as a console-script entry point; the importable package is `mesonbuild`. The old check was a false negative even with meson installed.
- **`build_numpy_ios.sh:28`** `PYHOST` now derives from `PYMOL_EXTERNAL_PREFIX` — consistent with the rest of the tree.
- **`build_ios.sh`** sets the `${PYMOL_EXTERNAL_PREFIX:-/opt/homebrew}` default before the `cmake` call and passes `-DPYMOL_EXTERNAL_PREFIX`, matching the CMake default.
- **`setup_ios_deps.sh`** references four scripts that all exist and are executable, in the right dependency order, under `set -euo pipefail`. `PYMOL_EXTERNAL_PREFIX=… scripts/setup_ios_deps.sh` propagates correctly to the children (inline `VAR=val cmd` exports it).
- **README commands** are accurate — checked each against `project.yml`: bundle id `io.raymol.RayMol` ✓, product/bundle `RayMol.app` (`PRODUCT_NAME: RayMol`) ✓, scheme `PyMOLViewer_iOS` ✓, project `PyMOLViewer.xcodeproj` ✓.

### Style / conventions

Clean and consistent with existing scripts (`#!/bin/bash`, `set -euo pipefail`, `ROOT="$(cd "$(dirname "$0")/.." && pwd)"`, absolute paths); new file committed executable (`100755`); comments explain *why*, matching the repo idiom. The README rewrite correctly frames physical-device signing as an Apple platform constraint, not a build-system gap.

### Minor suggestions (all non-blocking)

- **`setup_ios_deps.sh` runs the "optional" step-4 unconditionally.** With `set -e`, a Biopython download failure would abort the wrapper at 4/4 — *after* the expensive numpy cross-compile already succeeded. Consider `|| echo "step 4 (biopython, optional) failed — continuing"`, and echoing the effective `PYMOL_EXTERNAL_PREFIX` at the top for troubleshooting parity with `build_ios.sh`.
- **Precheck vs. what's actually used.** `build_numpy_ios.sh` checks `import mesonbuild, ninja` but the build uses numpy's *vendored* meson (line 43), and the check omits `cython` — which, per the new README note, is the thing that actually bites (console script not on `PATH`). Checking `cython` (or dropping the now-proxy meson check) would make the precheck match the documented failure mode.
- **`HOST_PYTHON` hint order (pre-existing, not in this diff).** `CMakeLists.txt:116` still lists `/opt/homebrew/bin` *before* `${PYMOL_EXTERNAL_PREFIX}/bin`, so a MacPorts machine that also has a Homebrew `python3` would pick the wrong interpreter for shader generation. Worth a follow-up to reorder the hints.
- **README simulator snippet** hardcodes `name=iPad Pro 11-inch (M5)` in the build command while the install snippet uses `<simulator UDID>` placeholders. A reader without that exact simulator gets "device not found" — a placeholder or "substitute your own" note would be more robust.

### Risk / test coverage

No runtime code (C++/Swift/Metal) changes — purely build scripts, CMake wiring, and docs, so there's no automated-test surface. The PR's own test plan is strong: both simulator and physical-device builds run end-to-end, `numpy.__version__` verified on-device.

**Verdict: Approve.** Correct, well-scoped, clearly documented; fixes two real bugs and materially improves iOS onboarding. Suggestions above are optional polish.

<sub>🤖 Reviewed with [Claude Code](https://claude.com/claude-code)</sub>

javierbq (approved):
ty!

PR #187, opened 2026-07-11
Stop shipping the classic Tk/Qt GUIs in the native app bundle

weitzner:
## Summary

First step of a longer-term effort to isolate a minimal PyMOL core that the
native Metal/SwiftUI apps actually depend on, separate from legacy classic-
desktop-GUI baggage carried over from the original PyMOL fork.

`pmg_tk`/`pmg_qt` (the classic Tk/Qt desktop GUIs) are confirmed unreachable
by the native app's actual build:
- `appkit/CMakeLists.txt` never references either.
- `layerGraphics/gl` (the classic OpenGL renderer) and the whole non-Metal
  AppKit desktop build path are excluded for both `PYMOL_IOS` and
  `PYMOL_METAL_ONLY` — the only two CMake configurations this app ships.

The adjacent bundling script already stripped the embedded Python's Tcl/Tk
*binaries* for exactly this reason (existing comment: "the native
Metal/SwiftUI app never launches the legacy pmg_tk Tkinter GUI, so the whole
Tcl/Tk stack is dead weight") — but the wholesale `ditto .../modules` copy
still shipped their Python *source* in every build regardless.

This only touches the app-bundling scripts (`project.yml` → regenerated
`project.pbxproj`) — nothing in the CMake build, nothing in `pip install .`,
nothing in the module source itself. Small, mechanical, easy to verify.

## Bigger picture (not in this PR)
Confirmed with Brian that the classic pip-installable Tk/Qt desktop
distribution is vestigial — not an actively shipped/supported RayMol product.
That opens up a larger cleanup opportunity (actually removing/archiving
`pmg_tk`, `pmg_qt`, `contrib/vr`, and the classic desktop OpenGL renderer
path from the tree, then reorganizing so the real "minimal PyMOL core" the
native apps depend on lives under one clear directory boundary, separate
from RayMol-original code). That's a bigger, shared-repo-visible decision
this PR intentionally doesn't make — happy to open a follow-up issue laying
out the reasoning if you want to discuss before anyone acts on it.

## Test plan
- [x] Regenerated `project.pbxproj` via `xcodegen generate` from the updated
      `project.yml`.
- [x] Rebuilt `PyMOLViewer_macOS` — succeeded.
- [x] Confirmed `modules/pmg_tk` and `modules/pmg_qt` are absent from the
      built `RayMol.app` bundle, while `modules/{pymol,chempy,pymol2,
      raymol_mcp}` remain present.
- [x] Launched the rebuilt app — runs normally.

### DIFF
--- a/swiftui/PyMOLViewer.xcodeproj/project.pbxproj
+++ b/swiftui/PyMOLViewer.xcodeproj/project.pbxproj
@@ -16,6 +16,7 @@
 		24169A74B9EBFD202258A5E4 /* PyMOLBridge.mm in Sources */ = {isa = PBXBuildFile; fileRef = 356FC561FA1515B3D15A26E2 /* PyMOLBridge.mm */; };
 		300656C47AD9A9C18C129C42 /* SequencePanel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 474D18A43D0D9A6BDD69EEFD /* SequencePanel.swift */; };
 		33C46A7EB6127A6A99D7048D /* TimelinePanel.swift in Sources */ = {isa = PBXBuildFile; fileRef = B820FE9080BF73526A54CD66 /* TimelinePanel.swift */; };
+		36D99340CCDDAD7C67680476 /* whatsnew-161-hover.mp4 in Resources */ = {isa = PBXBuildFile; fileRef = 6E2FC2AD1A8FEE912986DF9D /* whatsnew-161-hover.mp4 */; };
 		3A8107D094C920B600681B74 /* ThemeManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = EDC5E722722DEF7FBB1D649F /* ThemeManager.swift */; };
 		3D6FE0218915810960306973 /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 589928DEEE69075B965DB9F2 /* Assets.xcassets */; };
 		3E6C34959A3934D2DE3E080F /* RayMolUpdater.swift in Sources */ = {isa = PBXBuildFile; fileRef = F607D204A410CA38B84223BE /* RayMolUpdater.swift */; };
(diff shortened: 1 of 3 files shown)


javierbq:
### ⚠️ Heads-up: this breaks the release DMG build (`make_dmg.sh`)

The core change is sound — I verified the bundle boots via `PyMOL_Start`/`PInit` (never `launch()`, never `import pymol.plugins`) and that `pmg_tk`/`pmg_qt` were already unimportable at runtime (tkinter stripped, Qt not bundled), so removing their source is safe.

But there's one downstream impact the test plan didn't exercise: **`swiftui/make_dmg.sh` has a "package is not stale" guard that will now fail.**

`make_dmg.sh:161-178` runs, *before* signing/notarization:

```sh
diff -rq -x '__pycache__' -x '*.pyc' -x '.DS_Store' \
     "$PYMOL_ROOT/modules" "$BUNDLED_MODULES"   # exit 1 on ANY difference
```

This PR intentionally makes the bundle diverge from source: `pmg_tk`/`pmg_qt` stay in `modules/` (source) but are pruned from `Contents/Resources/modules` (bundle). Reproduced locally:

```
Only in modules: pmg_qt
Only in modules: pmg_tk
→ diff exits non-zero → "ERROR: ... the package is STALE" → exit 1
```

So the next Sparkle release cut (`cut-macos-release` → `make_dmg.sh`) will **fail at step "1c/8 Sanity"**. The dev `PyMOLViewer_macOS` build + launch (what the test plan covered) is unaffected — this only bites the DMG packaging path.

**Suggested fix, in lockstep with this PR** — teach the guard about the intentional pruning:

```sh
diff -rq -x '__pycache__' -x '*.pyc' -x '.DS_Store' -x pmg_tk -x pmg_qt \
     "$PYMOL_ROOT/modules" "$BUNDLED_MODULES"
```

(The MAS `archive_appstore.sh` flow has no equivalent diff guard and already prunes `raymol_mcp` the same way, so it's unaffected.)

Also worth a note in the PR body: the regenerated `project.pbxproj` picked up `whatsnew-161-hover.mp4` (a previously-unreferenced-but-on-disk resource referenced by `WhatsNew.json`) — a legit drift fix, but currently undocumented given the "only touches the app-bundling scripts" framing.

weitzner:
Good catch — fixed in de08eca8: `make_dmg.sh`'s staleness diff now excludes `pmg_tk`/`pmg_qt` (`-x pmg_tk -x pmg_qt`), matching the intentional bundle/source divergence.

Also rebased this branch onto current `master` (it was ~50 commits behind and had a real conflict in the generated `project.pbxproj`, resolved by regenerating via `xcodegen generate` from the cleanly-merged `project.yml` rather than hand-editing the generated file). That also means the `whatsnew-161-hover.mp4` pbxproj drift you flagged is no longer part of this PR's diff — it's already on `master`.

javierbq (approved):
LGTM!

PR #226, opened 2026-07-24
Load ~/.raymolrc startup script, migrated from ~/.pymolrc

### LINKED ISSUE #225: Support .raymolrc startup script (native apps skip .pymolrc entirely)
## Background

RayMol's native macOS/iOS apps never load a user startup script. Traced both embeddings:

- `pymol/__init__.py`'s `init2()` calls `invocation.parse_args(args)` (`modules/pymol/invocation.py`), which is what searches for `.pymolrc` / `.pymolrc.py` / `pymolrc.pml` and executes them. That path only runs in the standard `pymol` CLI (`layer5/main.cpp`).
- The legacy AppKit app (`layer5/main_appkit.mm`) and the current SwiftUI/Metal bridge (`swiftui/PyMOLViewer/Bridge/PyMOLBridge.mm`, see `PyMOLBridge_InitPython`/`PyMOLBridge_Start`) boot Python directly and wire up `PyMOL_Start`/`PInit` by hand — they never call `invocation.parse_args()`, so `.pymolrc` is silently ignored. A user's aliases/settings/startup commands from vanilla PyMOL have no effect in RayMol today.

## Proposal

1. Add a RayMol-native startup script, e.g. `~/.raymolrc` (and/or `~/.raymolrc.py`), loaded by both native embeddings after `PInit`/`PyMOL_SetPythonInitStage(..., 1)` — mirroring where `PyMOL_Idle`'s `exec_deferred()` runs so deferred commands work the same as CLI pymolrc.
2. First-run migration: the first time RayMol starts and detects an existing `~/.pymolrc` (or `.pymolrc.py`) but no `~/.raymolrc` yet, offer to import it — copy/rename it to `~/.raymolrc` (or wrap with a note) so existing PyMOL users' customizations carry over automatically instead of silently vanishing. Should only trigger once (track via a UserDefaults/settings flag or the presence of `~/.raymolrc` itself) and must not clobber a `.raymolrc` a user has already created.
3. Consider surfacing an "Edit raymolrc" menu item (parallel to the existing `edit_pymolrc` in `modules/pmg_qt/TextEditor.py` / `modules/pmg_tk/TextEditor.py`) for discoverability.

## Open questions

- Exact file name/location (`~/.raymolrc` vs an app-support-dir file) and whether iOS needs an equivalent (no home directory in the same sense — likely app-container-relative).
- Whether the import should be a straight copy, or executed once and then also saved as `.raymolrc` for future edits.
- Whether this should respect a `-k`-equivalent "skip startup script" flag/setting.

javierbq:
## Summary

Closes #225.

RayMol's native embeddings (SwiftUI/Metal and the legacy AppKit app) boot Python directly and never call `pymol.invocation.parse_args`, so `~/.pymolrc` is silently ignored there — a user's aliases/settings from vanilla PyMOL had no effect.

- `modules/pymol/raymolrc.py`: loads `~/.raymolrc(.py)` at startup. The first time neither raymolrc file exists yet, imports it from an existing `~/.pymolrc(.py)` so prior customizations carry over automatically (one-time, presence-of-file gated — never clobbers a `~/.raymolrc` the user has already edited). Also seeds `__main__` with `cmd`/`stored` before running the script, matching the vanilla CLI's namespace so bare `cmd.xxx()` calls (the common pymolrc convention) keep working — the native embeddings don't populate `__main__` the way the real `pymol` launcher script does.
- `swiftui/PyMOLViewer/Shared/ContentView.swift`: calls `raymolrc.load()` at the **end** of `applyPersistedTheme()`, not from the C++ bridge. VM testing caught a real ordering bug: the app's own launch-time theme reapply (bg color, etc.) runs asynchronously right after engine init and would otherwise silently overwrite anything the user's script set. Loading last means the user's script wins, matching how `.pymolrc` behaves in vanilla PyMOL (runs after built-in defaults).
- `swiftui/PyMOLViewer/Bridge/PyMOLBridge.mm`: removed the (incorrectly-ordered) load call from `PyMOLBridge_Start`, left a note pointing at the Swift-side call site.
- `layer5/main_appkit.mm`: same `raymolrc.load()` wiring for the legacy AppKit app, right after `PInit`/`PyMOL_SetPythonInitStage(1)` — this embedding has no theme-reapply step, so no ordering issue there.

## Test plan

Verified functionally in an isolated macOS VM (mac-vm-pool), driving the built `RayMol.app` over its MCP server:

- [x] **Migration**: with `~/.pymolrc.py` present (`cmd.set('bg_rgb', 'red')`, bare `cmd`, no import) and no `~/.raymolrc(.py)` yet, launching RayMol created `~/.raymolrc.py` with matching content and executed it — `cmd.get('bg_rgb')` returned `red` after launch, confirming it ran and survived the app's own theme reapply.
- [x] **Idempotency**: after editing `~/.raymolrc.py` to a different setting (`bg_rgb='blue'`) while leaving the original `~/.pymolrc.py` untouched, relaunching did **not** re-copy `~/.pymolrc.py` over the edited file — `~/.raymolrc.py` was preserved as edited, and its (new) content is what actually loaded (`bg_rgb` read back as `blue`).
- [x] `xcodebuild -scheme PyMOLViewer_macOS` succeeds (`-skipPackagePluginValidation -skipMacroValidation` for the mlx-swift SPM plugin trust gate).
- [x] `python3 -m py_compile modules/pymol/raymolrc.py`.

Not covered: iOS (no home-directory equivalent there — left for a follow-up per the open questions in #225), and an "Edit raymolrc" menu item (also left as a follow-up).

### DIFF
--- a/layer5/main_appkit.mm
+++ b/layer5/main_appkit.mm
@@ -363,6 +363,17 @@ static void handleKeyDown(NSView *view, NSEvent *event) {
     // which calls cmd.config_mouse() to set up mouse bindings.
     PyMOL_SetPythonInitStage(pymolInstance, 1);
 
+    // Load ~/.raymolrc(.py), importing it from ~/.pymolrc(.py) on first
+    // detection — this embedding never goes through pymol.invocation's CLI
+    // argument parsing, so vanilla PyMOL's .pymolrc is otherwise never read
+    // (RayMol#225).
+    PyRun_SimpleString(
+        "try:\n"
+        "    from pymol import raymolrc as _raymolrc\n"
+        "    _raymolrc.load()\n"
+        "except Exception as _e:\n"
+        "    import os; os.write(2, ('[PyMOL] raymolrc load failed: %r\\n' % (_e,)).encode())\n");
+
     // Load API keys into Python's os.environ before importing ai_chat.
     // Finder-launched apps don't inherit shell env vars, so we also
     // check ~/.pymol_ai.conf (simple KEY=VALUE format).
--- a/layer5/main_appkit.mm
+++ b/layer5/main_appkit.mm
@@ -363,16 +363,52 @@ static void handleKeyDown(NSView *view, NSEvent *event) {
     // which calls cmd.config_mouse() to set up mouse bindings.
     PyMOL_SetPythonInitStage(pymolInstance, 1);
 
-    // Load ~/.raymolrc(.py), importing it from ~/.pymolrc(.py) on first
-    // detection — this embedding never goes through pymol.invocation's CLI
-    // argument parsing, so vanilla PyMOL's .pymolrc is otherwise never read
-    // (RayMol#225).
-    PyRun_SimpleString(
-        "try:\n"
-        "    from pymol import raymolrc as _raymolrc\n"
-        "    _raymolrc.load()\n"
-        "except Exception as _e:\n"
-        "    import os; os.write(2, ('[PyMOL] raymolrc load failed: %r\\n' % (_e,)).encode())\n");
+    // Load ~/.raymolrc(.py). This embedding never goes through
+    // pymol.invocation's CLI argument parsing, so vanilla PyMOL's .pymolrc is
+    // otherwise never read (RayMol#225). The first time no ~/.raymolrc(.py)
+    // exists yet and a ~/.pymolrc(.py) is found, ask before importing it
+    // rather than copying it silently.
+    {
+        NSFileManager *fm = [NSFileManager defaultManager];
+        NSString *home = NSHomeDirectory();
+        BOOL hasRaymolrc = [fm fileExistsAtPath:[home stringByAppendingPathComponent:@".raymolrc.py"]]
+            || [fm fileExistsAtPath:[home stringByAppendingPathComponent:@".raymolrc"]];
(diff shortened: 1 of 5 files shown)


javierbq:
Updated: the first-run migration is no longer a silent copy. Both native apps now show a confirmation prompt the first time an unmigrated `~/.pymolrc(.py)` is found:

> **Import your PyMOL startup script?**
> RayMol found an existing ~/.pymolrc and can copy it to ~/.raymolrc, RayMol's own startup script, so your customizations still run here.
> [Not Now]  [Import]

- **Import** → copies it to `~/.raymolrc(.py)` and runs it immediately.
- **Not Now** → writes `~/.raymolrc.skip` so it won't ask again, and does not touch `~/.pymolrc`.

`modules/pymol/raymolrc.py` now splits the concerns: `load()` just runs whichever `~/.raymolrc(.py)` files exist (no migration side effects), while `migration_candidate()` / `migrate()` / `decline_migration()` are driven by the native UI — a SwiftUI `.alert` on macOS/iOS, an `NSAlert` in the legacy AppKit app.

VM-verified all four paths (prompt fires instead of auto-copying, Import migrates + executes, Not Now declines cleanly, second launch doesn't re-prompt).

PR #229, opened 2026-07-24
Scenes: capture per-object TTT matrices (Move-mode position/orientation) (#204)

### LINKED ISSUE #204: Scenes: capture per-object TTT matrices (Move-mode position/orientation)
**Is your feature request related to a problem? Please describe.**

Scenes don't remember where objects have been moved. With **Move mode** (new in 1.7.0, *Mouse ▸ Move Objects* / ⌃M) you can rigid-body translate/rotate individual objects. But if you then store a scene, the object's position/orientation is **not** captured — recalling the scene restores the camera, representations, colors, and render settings, yet leaves each object wherever it currently sits. Building a set of scenes (or a movie from them) that show objects in different arrangements is therefore impossible today.

**Root cause**

- Move mode applies the transform as a per-object **TTT matrix** in the core (`cmd.translate`/`cmd.rotate ... object=`, see `modules/pymol/metal_move.py`). This is non-destructive and *does* persist in a saved `.pse`.
- Classic PyMOL `scene store`/`recall` only captures camera view + reps + colors — **not** per-object TTT matrices. `modules/pymol/raymol_scenes.py` already extends scenes to also snapshot render settings (the same class of "scene doesn't capture this" gap), but it does not touch object matrices.

So the moved layout survives Save Session but not scene recall.

**Describe the solution you'd like**

Extend RayMol's scene capture to include each object's TTT matrix, mirroring the pattern `raymol_scenes.py` already uses for render settings:

- On `scene ... store` / `scene ... update` (`snapshot_current`): record `cmd.get_object_ttt(obj)` (or `get_object_matrix`) for every current object, keyed by scene name.
- On `scene ... recall` / `apply`: re-apply via `cmd.set_object_ttt(obj, ...)` for objects still present; skip objects that no longer exist.
- Persist the per-scene TTT map through the existing `session_save` / `session_restore` hooks so it round-trips in `.pse`.

**Acceptance criteria**

- Move an object, `scene A, store`; move it elsewhere, `scene B, store`. Recalling A vs B moves the object between the two positions.
- The per-scene positions survive a `.pse` save/reload.
- A movie built from scenes animates the object positions (interpolated or stepped — see open questions).
- Objects added/removed between store and recall don't error; missing objects are skipped.

**Describe alternatives you've considered**

- Document the limitation and tell users to Save Session instead — insufficient, since scenes/movies are the natural workflow for arrangements.
- Bake moves into coordinates (`matrix` apply) so they're captured as structure — rejected: destructive, defeats the non-destructive design of Move mode.

**Additional context**

- Discovered while testing the 1.7.0 RC (Move mode is the headline feature); this is expected long-standing PyMOL scene behavior, not a 1.7.0 regression, so it's a follow-up rather than a release blocker.
- Open questions: should timeline/movie transitions **interpolate** object TTT between scene keyframes (smooth object motion) or step? Interaction with multi-state objects and per-state matrices. Whether to gate behind a scene-content flag for vanilla-parity.
- Relevant files: `modules/pymol/raymol_scenes.py`, `modules/pymol/metal_move.py`, scene recall path in `swiftui/PyMOLViewer/Shared/PyMOLEngine.swift`.

javierbq:
Closes #204.

Scenes now capture each object's **Move-mode TTT** (rigid-body position/orientation), restore it on recall, persist it in `.pse`, and **interpolate it smoothly in movies** built from scenes.

## Approach

Investigation surfaced two facts that reshaped the issue's original "just mirror `raymol_scenes.py`" framing:

- **The hook-pairing pattern was fragile.** `cmd.scene()` never called `raymol_scenes`; only the ObjectPanel Scenes-tab buttons paired it manually. The viewport overlay, timeline menus, **rename** (orphaned snapshots), console, and MCP/agent all bypassed it — latent bugs the shipped render-settings feature shared.
- **Movie playback recalls scenes in C++** and never fires a Python hook, so a per-recall callback can't animate a movie. But the core natively interpolates per-object TTT if `cmd.mview(object=…)` keyframes are authored at build time.

So this PR:

1. **Centralizes** per-scene extras (new object TTT **and** the existing render settings) behind a single guarded hook in `cmd.scene` (`viewing.py`), firing on every scene action from any path (UI/console/MCP/movie-assembly). This also fixes the latent rename/overlay/console/MCP gaps for the render-settings feature.
2. **Captures** `get_object_ttt` per object (identity for unmoved, so recall can also *reset*), **restores** via `set_object_ttt`, skipping absent objects.
3. **Guards** internal temp-scene machinery (`.pse` legacy restore, legacy export) with a `suspended()` context so it doesn't wipe/pollute the store.
4. **Authors** interpolated object-motion keyframes in all scene-movie build paths (`appkit_movie.rebuild` / `append_template` / `place_scene`, and classic `movie.add_scenes` incl. holding through the camera dwell).
5. **Persists** a `raymol_scene_ttt` map through the existing `.pse` session tasks (tolerates old files).
6. Removes the now-redundant Swift `_rs.*` pairings.

## Capture API

Capture uses `get_object_ttt` / restore uses `set_object_ttt` (isolates the TTT channel; avoids double-counting an object's state matrix, which `get_object_matrix` would). A test does an off-center rotation — the exact case that diverges the Move gizmo — and asserts a byte-identical round-trip.

## Testing

- **Headless pytest** (`testing/tests/test_raymol_scene_ttt.py`, 9 tests): capture/restore incl. off-center rotation, reset-to-unmoved, add/remove tolerance, rename rekey, dispatcher, `suspended()`, `emit_object_motion`, session round-trip.
- **PyMOLTestCase** (`testing/tests/raymol/scene_ttt.py`) for the RayMol `--testing` CI build: the `cmd.scene` hook end-to-end, `.pse` save/load, rename, and movie motion (verified by ray-rendered frame differences — `get_object_matrix` after `cmd.frame` does *not* reflect movie object motion, which is applied in the render path).
- **Functional**, on a real macOS build driven over MCP in a disposable VM: manual recall jumps the object between stored positions; `.pse` round-trips; a scenes movie **glides** the object smoothly (frame 1 centered → midpoint intermediate → end translated+rotated).

## Follow-ups (non-blocking)

- Automated coverage smoke for the `append_template`/`place_scene`/`add_scenes` authoring paths (only `rebuild` has an automated motion test; others covered by the functional pass).
- Object-motion + per-object state-sweep coexistence in one movie (channels are architecturally independent).

Design + plan: `docs/superpowers/specs/2026-07-24-scene-object-ttt-capture-design.md`, `docs/superpowers/plans/2026-07-24-scene-object-ttt-capture.md`.

### DIFF
--- a/docs/superpowers/plans/2026-07-24-scene-object-ttt-capture.md
+++ b/docs/superpowers/plans/2026-07-24-scene-object-ttt-capture.md
@@ -1,0 +1,1005 @@
+# Per-Scene Object TTT Capture Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Scenes capture each object's Move-mode TTT (position/orientation), restore it on recall, persist it in `.pse`, and smoothly animate it in movies built from scenes.
+
+**Architecture:** `raymol_scenes.py` becomes the unified per-scene "extras" backend (render settings + object TTT). A single guarded hook in `cmd.scene` (`viewing.py`) fires capture/restore on every scene action from any path (UI, console, MCP, movie-assembly). Movie object motion is authored by emitting `cmd.mview('store', object=…)` keyframes at build time, which the core interpolates during playback and export.
+
+**Tech Stack:** Python (PyMOL `cmd` API, `pymol2` headless for tests), SwiftUI (removal of now-redundant calls), the RayMol `--testing` build + a disposable macOS VM for functional verification.
+
+## Global Constraints
+
+- Python 3.9+ (repo floor). No Python formatter configured — match surrounding style.
+- Every new cross-module dependency on `raymol_scenes` must be **import-guarded** (try/except → no-op) so upstream merges and non-RayMol builds stay green. This mirrors the existing guarded registration in `modules/pymol/cmd.py:59-66`.
+- Capture uses `cmd.get_object_ttt` / restore uses `cmd.set_object_ttt` (isolates the TTT channel; do **not** use `get_object_matrix` for capture — it bakes in the state matrix and won't round-trip through `set_object_ttt`). `get_object_ttt` returns `None` for an unmoved object; substitute identity on restore.
+- Identity TTT (16-float): `[1,0,0,0, 0,1,0,0, 0,0,1,0, 0,0,0,1]`.
+- Functional/UI verification MUST run in the mac VM (or iOS sim) via the `raymol-mac-vm` / `mac-vm-test` skill — never drive the host UI (per CLAUDE.md + user directive).
+- Do not commit to `master`. Work stays on branch `claude/issue-204-0609e7`; open a PR at the end.
+- Bare-pytest tests use the repo venv: `/Users/jcastellanos/repos/RayMol/.venv/bin/python`. The `PyMOLTestCase` runner (`pymol -ckqy testing/testing.py --run …`) only works on a RayMol `--testing=True` build (the venv/Homebrew pymol lack `pymol.testing`).
+
+---
+
+## File Structure
+
+- `modules/pymol/raymol_scenes.py` (modify) — add `_scene_ttt` store, TTT capture/restore, `rename`, `on_scene_action` dispatcher, `suspended()` guard, `scene_ttt_map`, `emit_object_motion`, and extend session save/restore. Single responsibility: per-scene extras storage + apply.
+- `modules/pymol/viewing.py` (modify) — call the hook at the end of `cmd.scene`; wrap `session_restore_scenes` temp-scene ops in `suspended()`.
+- `modules/pymol/exporting.py` (modify) — wrap the legacy-scene temp block in `get_session` in `suspended()`.
+- `modules/pymol/appkit_movie.py` (modify) — emit object-motion keyframes in `rebuild`, `append_template` (scenes), `place_scene`.
+- `modules/pymol/movie.py` (modify) — emit object-motion keyframes in `add_scenes`.
+- `swiftui/PyMOLViewer/Panels/ObjectPanel.swift` (modify) — remove the now-redundant `_rs.*` pairing calls.
+- `testing/tests/test_raymol_scene_ttt.py` (create) — bare-pytest backend tests (runs locally now).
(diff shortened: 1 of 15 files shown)


javierbq:
## Follow-up: per-scene render settings now animate in scene movies

Addresses the report that *"scenes now save the DOF setting, but when making a video only the last active setting is used."*

**Root cause:** movie playback recalls scenes entirely in C++ (`MovieDoFrameCommand` → `MovieSceneRecall`) and never fires the Python `cmd.scene` hook, so the per-scene render settings were never applied during a movie. `ViewElem` has no channel for global settings, so the only vehicle is a per-frame movie command.

**What landed** — a new `modules/pymol/raymol_scene_anim.py`, wired into all four scene-movie authoring paths:

- Continuous settings (DOF focus/range/aperture, exposure, lighting, outline width, AO…) **interpolate** across each transition using a branch-exact port of the core's easing (`View.cpp:878-897`), so they track the camera instead of desyncing. Discrete settings and quality knobs step at the cut.
- Differing Auto-lock targets produce a true **focus pull**: autofocus is switched off across the transition and the distance is driven per frame by reprojecting the interpolated target centroid under that frame's camera (a straight endpoint lerp drifts when the camera dollies).
- The track is persisted as **structured data** and the command strings are regenerated locally on restore. This is deliberate: any raw frame command left in a `.pse` trips `MovieSetLock`, which disables the whole per-frame path — commands, scene recall **and** the camera track — and RayMol has no unlock UI.

**Verified on the compiled macOS build in a disposable VM:** aperture ramps `1.0 → 2.516 → 5.0 → 7.484 → 9.0` across a 21-frame two-scene movie (2.516 at t=0.25 confirms the eased curve; a linear ramp would be 3.0); Linear vs Smooth transitions now differ correctly; the focus pull is monotone with autofocus re-locked at each keyframe; and a `.pse` round-trip keeps both the camera track and the animation live. 49 headless tests.

**Review found and fixed** two `.pse`-safety defects (re-authoring orphaned commands → locked movie; a stale track leaking into an unrelated movie), two model mismatches against the core (easing resolved from only one endpoint; the focus centroid using an untransformed mean instead of the renderer's transformed bbox midpoint, which also fought the per-object TTT feature), and two vacuous tests. It also disproved a limitation I had written into the spec — `add_scenes`' `_rock`/`_nutate` author `mview` keyframes, not frame commands — so that caveat is gone rather than shipped.

**Known gap (deferred):** movie *export* shares the verified `cmd.frame` path, but I verified value-by-frame rather than by inspecting an exported file.

PR #242, opened 2026-08-03
Add session-linked Analysis Notes workflow

pgreisen-16:
## Summary

Adds a session-linked Analysis Notes workspace to RayMol so structural observations stay beside the molecular context that produced them.

## What changed

- adds an Analysis Notes inspector with multiple named Markdown note documents
- embeds notes, metadata, and linked images directly in the `.pse` session; legacy sidecars remain readable for migration
- adds camera-only and full-scene view links that restore the corresponding PyMOL view
- makes inserted residue identifiers clickable so they select and zoom to the residue in 3D
- adds structured inserts for selected residues, nearby contacts/interfaces, and current measurements
- adds note search, Markdown heading outline, lightweight tags, scientific templates, and text-size controls
- supports native Apple keyboard Dictation without app-level microphone permissions, with platform-appropriate guidance
- adds Edit/Preview, view-link, and font-size keyboard shortcuts
- captures Metal-rendered screenshots and keeps them linked to the note
- exports clean Markdown, self-contained HTML, and paginated PDF with embedded images
- shares one self-contained `.pse` instead of sidecar and image companion files
- adds a detachable Analysis Notes window on macOS and a dedicated compact iPhone workspace
- keeps Insert View Link prominent in narrow layouts and prevents object-panel state collisions between multiple RayMol instances
- documents the workflow and adds focused persistence/export tests

## Why

RayMol users often need to record hypotheses, residue-level observations, measurements, and visual states while inspecting a structure. Keeping those notes inside the application reduces context switching and makes an analysis reproducible when a session is reopened or shared.

## Review feedback addressed

- stores the Notes payload in PyMOL's extensible session dictionary
- stores image assets as base64 data keyed by MD5 and materializes them on load
- makes residue insertion harmless when no selection exists
- improves narrow-panel spacing, search visibility, and control discoverability
- removes misleading custom microphone behavior while retaining native system Dictation

## Validation

- signed macOS Debug build succeeded
- focused `AnalysisNotesStoreTests`: 9 passed, 0 failed
- Analysis Notes Python session stage/save/restore/export round-trip passed
- Python syntax validation passed for the new and modified PyMOL modules
- macOS and iOS Swift frontend parsing passed for the updated Notes UI
- `git diff --check` passed


docs/analysis-notes.md
```diff
@@ -0,0 +1,56 @@
...
+
+The export menu supports clean Markdown, self-contained HTML with base64-embedded screenshots, and paginated PDF with images, headings, page numbers, and captions.
+
+On macOS, choose the window button in the Notes footer to open a detachable **Analysis Notes** window. It stays synchronized with the Notes inspector and can remain beside the molecular viewport.
+
+## Saving and sharing
+
+For `experiment.pse`, RayMol writes:
+
+- `experiment.raymol-notes.json` — Markdown, view-link metadata, and image metadata.
```
javierbq: I don't like that we are generating another set of files that need to be moved around with the pse.  A pse file is a serialized dictionary; we can append new keys and extend the pse format in a way that is backward compatible (PyMOL ignores extra fields). Instead of a set of companion files, the data of the notes should be stored under a new key in sessions. I think the content can be saved as a simple Markdown string and the assets(images) in a nested dict with MD5 digests as keys and binary content of the image as base64-encoded values. When a new session is loaded, the contents are laid out into a temp folder for the markdown references to work as expected. when a new unsaved session has been created a tmp folder should be made to work as the staging place until the session is saved. Changes should be automatically saved to the temp folder; then, when the PSE session is saved, the contents of the temp folder should be injected into the session state before writing the file.

swiftui/PyMOLViewer/Shared/PyMOLEngine.swift
```diff
@@ -1952,6 +1971,64 @@ final class PyMOLEngine: ObservableObject {
...
+    /// outside the current `sele` selection with an atom within the cutoff.
+    func noteResidues(contacts: Bool, cutoff: Double = 4.0) -> [[String: String]] {
+        guard isReady else { return [] }
+        let url = FileManager.default.temporaryDirectory
+            .appendingPathComponent("raymol-note-residues-\(UUID().uuidString).json")
+        let encodedPath = Data(url.path.utf8).base64EncodedString()
+        let selection = contacts
+            ? "byres (((sele) around \(cutoff)) and not (sele))"
+            : "sele"
+        runPython("""
```
javierbq, line 2043: throws error if no selection has been made but very cool feature how the selections are link themselves.

javierbq (changes requested):
Image This menus are not rendering properly, chevron is on top of icon. 

Image 

This indicator doen't says much, I think it should be removed to live more space for the buttons.

Image
search hint is not displayed properly

Image
buttons should hug tightly the content.

swiftui/PyMOLViewer/Panels/NotesInspectorView.swift
```diff
@@ -0,0 +1,1252 @@
...
+    private let debounceInterval: TimeInterval
+    private var pendingSave: DispatchWorkItem?
+    private var isLoading = false
+    private var hasOpenedSession = false
+    private var stageEmbeddedDocument: ((URL, URL) -> Bool)?
+    private var exportEmbeddedDocument: ((URL, URL) -> Bool)?
+
+    init(fileManager: FileManager = .default,
+         fallbackDirectory: URL? = nil,
+         debounceInterval: TimeInterval = 0.6) {
```
javierbq: debounce is to small. autosave is halting the UI and slowing down the rendering of what is being typed because calls are pilling up. increase to 10s.

swiftui/PyMOLViewer/Panels/NotesInspectorView.swift
```diff
@@ -0,0 +1,1252 @@
...
+        try? FileManager.default.removeItem(at: url)
+        isPreviewing = true
+    }
+
+    private func insertResidueSummary(contacts: Bool) {
+        let rows = engine.noteResidues(contacts: contacts)
+        let heading = contacts ? "## Contacts around selection (4.0 Å)" : "## Selected residues"
+        var lines = [heading, ""]
+        if rows.isEmpty {
+            lines.append("_No residues found. Create a `sele` selection in the 3D view first._")
```
javierbq: This should open a popup instead of polluting the notes

swiftui/PyMOLViewer/Panels/NotesInspectorView.swift
```diff
@@ -0,0 +1,1252 @@
...
+        }
+        appendMarkdown(lines.joined(separator: "\n") + "\n")
+        isPreviewing = true
+    }
+
+    private func insertMeasurementSummary() {
+        let rows = engine.noteMeasurements()
+        var lines = ["## Measurements", ""]
+        if rows.isEmpty {
+            lines.append("_No RayMol measurements have been created in this run._")
```
javierbq: Popup istead

swiftui/PyMOLViewer/Panels/NotesInspectorView.swift
```diff
@@ -0,0 +1,1252 @@
...
+        #if os(macOS)
+        guard let image = NSImage(contentsOf: url) else { return nil }
+        return image.cgImage(forProposedRect: nil, context: nil, hints: nil)
+        #else
+        return UIImage(contentsOfFile: url.path)?.cgImage
+        #endif
+    }
+}
+
+private struct AnalysisNoteTemplate: Identifiable {
```
javierbq: From the UI it's not obvious if this are templates to be introduced or that they are action that produces a calculation. I prefer we remove this templates for clarity

swiftui/PyMOLViewer/Panels/NotesInspectorView.swift
```diff
@@ -0,0 +1,1252 @@
...
+                Button { useSystemDictation() } label: {
+                    Image(systemName: "keyboard")
+                }
+                .help(dictationHelpText)
+
+                Menu {
+                    Button("Export Clean Markdown…") { showingMarkdownExporter = true }
+                    Button("Export HTML with Images…") { showingHTMLExporter = true }
+                    Button("Export PDF with Images…") { showingPDFExporter = true }
+                    Button("Sharing Notes & Sessions…") { showingShareHelp = true }
```
javierbq: Remove, this open and information modal instead of action. inconsistent with the rest of the actions.

swiftui/PyMOLViewer/Panels/NotesInspectorView.swift
```diff
@@ -0,0 +1,1252 @@
...
+                .help("Insert a Metal-rendered screenshot")
+
+                Button { useSystemDictation() } label: {
+                    Image(systemName: "keyboard")
+                }
+                .help(dictationHelpText)
+
+                Menu {
+                    Button("Export Clean Markdown…") { showingMarkdownExporter = true }
+                    Button("Export HTML with Images…") { showingHTMLExporter = true }
```
javierbq: Functionality verified but only the export pdf ask for target folder for saving assets.

javierbq (changes requested):
Save mechanism is almost perfect. There are two bugs related to autosave and file paths (check comments left on code)

other bugs: 

- when restoring a pse session tmp assets are written to the same folder as the pse, they should be in a tempory folder.
- Two camera icons, it should be one with insert image/camera/scene
- remove templates from bottom bar
- remove the view/tab counter indicator
- fix the position of the chevron in the buttons so they don't hide the icon in the button.
- add on hover tooltip to buttons
Image

pgreisen-16:
The requested autosave, temporary-file, insertion-popup, export, and toolbar changes are included in commit f66867f6c. All nine focused Analysis Notes tests pass, and the updated view parses successfully for both macOS and iOS.



4:16 PM

javierbq (changes requested):
Image. 

Image

Preview is not rendering the linked image in the right place. 

Image Remove the dictation button because it's not doing anything, just opening an informative message about how it doesn't work.

javierbq:
Re-tested on an iPhone 17 Pro simulator and a clean macOS VM.

### Both items I raised are fixed ✅

- **Preview image placement** — a note of `Foo` / `Molecular view 1` / `Bar` now renders text → image → text, interleaved where the marker sits, with no `LINKED IMAGES` block at the bottom. Fixed by 648e1876b. (My earlier report was measured against a build that predated that commit — sorry for the noise.)
- **Full-screen Notes on iPhone** — Notes now stays docked in the inspector with the viewport visible, instead of taking over the screen. Fixed by 41c12dc79. Thanks for the quick turnaround on both.

### I pushed de61e3077 to this branch

Three iOS-only defects I hit while testing the panel. Shout if you'd rather I'd filed them instead of pushing.

**1. The in-app Open pill never rebound the notes store.** `iosHandleImport` accepts `.pse` but never set `engine.currentSessionURL`, which is the only trigger for `notes.openSession`. So reopening a session kept the *previous* note on screen, and the next `persist()` staged that stale note back over the payload the opened `.pse` had just restored — meaning the following save wrote the wrong notes into the file. Now mirrors `macOpenFile` / `loadOpenedFile`, which both publish the URL after the structure loads.

**2. "Unsaved session" was misleading on iOS.** The label is a document-binding indicator (`sessionURL == nil`), not a dirty flag — but iOS has no document model: `iosSaveSession` and `iosShareSession` both export a *copy* (document picker / activity sheet), so the app never holds a writable URL. The label was therefore shown almost always, including right after the user saved and after a session was restored from autosave. It reads as "your work is at risk", which is the opposite of the truth, since notes ride inside the autosave `.pse`. It's now shown only when it can name a real document. macOS is unchanged.

I first tried binding it in the save/share paths — that doesn't work: `UIActivityViewController` has no destination URL at all, and the picker's `asCopy: true` export gives no ongoing write access, so binding would show a filename the app can never save back to.

**3. A notes-only session was silently destroyed.** `autosaveSession()` cleared the autosave for an object-less scene. Repro on a simulator before the fix: open the app, type a note *without loading a structure*, background it, relaunch → the note is gone, with no recovery path. It's now kept when the note has content.

Verified on device sim: that exact sequence now writes a 1.7 KB `autosave.pse` with the note embedded and the text is back in the editor after a cold relaunch. `AnalysisNotesStoreTests` stays 11/11, and both `PyMOLViewer_iOS` and `PyMOLViewer_macOS` compile (worth remembering that nothing in CI compiles the iOS target).

### Left for you — the root cause behind #3

**The store's own recovery file can never be read back.** f66867f6c moved it from Application Support to `{tmp}/RayMol/AnalysisNotes/{getpid()}/untitled.json` ([NotesInspectorView.swift#L100-L106](https://github.com/javierbq/RayMol/blob/de61e3077/swiftui/PyMOLViewer/Panels/NotesInspectorView.swift#L100-L106)). Because the path is keyed by PID, a new process computes a fresh empty directory — the previous launch's file is unreachable, and the old directories accumulate with nothing cleaning them up. I watched a note sit intact on disk under a dead PID while the app showed an empty editor.

That leaves the `.pse` as the only durable home a note has on iOS, which is what made #3 destructive. I didn't touch it because the PID scoping looks deliberate for multi-instance isolation, and the right fix is your call — presumably a stable per-session key in Application Support, with the instance discriminator only where two RayMols could actually collide. Note the engine already reasons this way for the autosave: *"A subfolder of Library survives across launches (unlike tmp, which iOS may purge)"* ([PyMOLEngine.swift#L665-L667](https://github.com/javierbq/RayMol/blob/de61e3077/swiftui/PyMOLViewer/Shared/PyMOLEngine.swift#L665-L667)).

Smaller things I noticed but left alone:

- `openSession` calls `flush()` **before** `loadEmbeddedDocument`, so a dirty note can stage itself over the payload of the `.pse` you just opened, and the load then reads it straight back as if it belonged to that file.
- `stageAnalysisNotes` returns `true` unconditionally after only an `isReady` + file-exists check, and a raise inside `raymol_notes.stage` is merely `PyErr_Print()`ed. Combined with `saveStatus` rendering nothing for both `.saved` and `.pending`, a Python-side staging failure produces a `.pse` with no notes and zero UI signal.
- `selectPage()` doesn't set `.pending` or call `scheduleSave()`, unlike its neighbours, so a page switch alone is never staged.
- Every `AnalysisNotesStoreTests` case injects an explicit `fallbackDirectory` and stubs `export`, so the shipped default path and the embed/restore round-trip aren't covered. A test that exercises the real default would have caught #3.

PR #251, opened 2026-08-05
Fix #219: skip transparency probes for non-molecular objects

### LINKED ISSUE #219: Creating a measurement spams the console with "Invalid selection name \"dist01\"" (object poll runs cmd.iterate on non-molecular objects)
## Summary

After creating a **measurement** (Distance / Angle / Dihedral), the console falls into an endless error loop, repeating roughly every ~500 ms for as long as the measurement object exists:

```
dist01<--
 Selector-Error: Invalid selection name "dist01".
dist01<--
 Selector-Error: Invalid selection name "dist01".
...
```

Nothing is functionally broken (no crash, the measurement itself works), but the feedback log is flooded and becomes unusable — the same class of console-spam bug as #113 and #145.

## Steps to reproduce

1. Load a structure (e.g. `fetch 2kpo`).
2. Open the **Measure** tool and take a **Distance** between two atoms. This creates a distance object named `dist01`.
3. Watch the console.

**Actual:** `Selector-Error: Invalid selection name "dist01".` repeats indefinitely (~every 500 ms).
**Expected:** no console output after the measurement is created.

Same happens for angle (`ang01`) and dihedral objects, and for any non-molecular object (CGO, etc.).

## Root cause

The Objects-panel poll probes **every** public object for a per-atom-transparency override, without filtering by object type.

- `PyMOLEngine.pollObjects()` runs every ~500 ms and evaluates, for *all* objects:
  `swiftui/PyMOLViewer/Shared/PyMOLEngine.swift:2474`
  ```python
  _ht = {o: _ai.object_has_atom_transp(o) for o in _objs}   # _objs = get_names('public_objects')
  ```
- `object_has_atom_transp(obj)` → `transp_summary(obj)` calls:
  `modules/pymol/appkit_inspector.py:188`
  ```python
  cmd.iterate(obj, expr, space={'_visit': _visit})
  ```
- When `obj` is a **distance object** (`ObjectDist`, created at `modules/pymol/appkit_measure.py:79` via `cmd.distance('dist%02d' % _counter, ...)`), `cmd.iterate` tries to parse `dist01` as an *atom selection*. A measurement object is not a molecular object, so the C++ selector rejects it and emits `Selector-Error: Invalid selection name "dist01".` to the feedback log.
- The `try/except` around the `iterate` (appkit_inspector.py:187–190) swallows the *Python* exception (hence no traceback / no crash), but the selector's error line was already written to the feedback buffer before the exception was raised — so it can't be suppressed that way. Result: pure log spam, once per poll, forever.

The expanded-card detail path (`_build` / `poll`, `appkit_inspector.py:218`+) has the same issue if a measurement object's card is expanded.

## Suggested fix

Restrict the per-atom-transparency probe (and the expanded-card `_build`) to molecular objects, e.g. gate on `cmd.get_type(o) == 'object:molecule'` before calling `object_has_atom_transp(o)` / `transp_summary(o)`, or filter `_objs` in the poll query. Non-molecular objects can't carry per-atom transparency anyway, so skipping them is both correct and cheaper.

## Environment

- RayMol native app (shared `PyMOLEngine` + bundled `appkit_inspector`), current `master`.
- Affects both macOS and iOS/iPadOS (the poll and inspector code are in the shared layer).

GhagSagar23:
## Summary

- Skip per-atom transparency inspection for measurement, CGO, and other non-molecular objects.
- Apply the guard centrally so it covers both periodic object-panel polling and expanded object cards.
- Preserve `has_transp` entries as `false` for non-molecular objects.
- Add focused regression coverage for molecule, distance, angle, dihedral, and CGO objects.

## Problem

The object panel probes every public object for per-atom transparency. `transp_summary()` passed each object name to `cmd.iterate()`, which treats the name as an atom selection. Measurement and other non-molecular object names are not valid atom selections, so the selector wrote an error to the feedback log before the Python exception handler could suppress it. Because the panel polls repeatedly, this flooded the console.

## Solution

Return an empty transparency summary when `cmd.get_type(obj)` is not `object:molecule`. Non-molecular objects cannot carry per-atom transparency overrides, and placing the guard in `transp_summary()` protects both the collapsed-row polling path and the independently reachable expanded-card path.

The regression test builds a mixed session and verifies that only the molecular object reaches `cmd.iterate()`. It also checks that non-molecular objects retain false transparency flags and empty expanded-card details.

## Testing

- PASS: `git diff --check`
- PASS: changed-file secret-pattern scan
- NOT RUN: focused PyMOL object-panel tests; no reviewed compatible isolated runner was available.
- NOT RUN: broader PyMOL test suite.
- NOT RUN: macOS and iOS Xcode builds.
- NOT RUN: native smoke verification across repeated poll intervals and expanded object cards.

Runtime behavior therefore still needs verification in a reviewed isolated macOS environment.

Fixes #219

### DIFF
--- a/modules/pymol/appkit_inspector.py
+++ b/modules/pymol/appkit_inspector.py
@@ -171,6 +171,13 @@ def transp_summary(obj):
     no atom-level override exists (it never returns None here), so comparing the
     effective range to the object-level value is what detects a genuine override.
     """
+    # Measurement, CGO, map, and other non-molecular objects are not atom
+    # selections. Passing their names to iterate emits a Selector-Error before
+    # Python can catch the exception, which floods the native app's feedback log
+    # on every object-panel poll. They cannot carry per-atom overrides anyway.
+    if cmd.get_type(obj) != 'object:molecule':
+        return {}
+
     objlv = {s: _num(s, obj) for s in TRANSP_SETTINGS}
     mn = {s: None for s in TRANSP_SETTINGS}
     mx = {s: None for s in TRANSP_SETTINGS}
--- a/testing/tests/test_appkit_objpanel_poll.py
+++ b/testing/tests/test_appkit_objpanel_poll.py
@@ -23,8 +23,9 @@ import io
 import json
 import os
 import tempfile
+from unittest.mock import patch
 
-from pymol import cmd, testing
+from pymol import cgo, cmd, testing
 from pymol import appkit_inspector as ai
 
 # layer0/PyMOLGlobals.h: OrthoLineLength — the per-line feedback cap that the
@@ -72,6 +73,21 @@ class TestObjPanelPoll(testing.PyMOLTestCase):
                 cmd.select(sel, '%s and (%s)' % (obj, expr), quiet=1)
                 sels.append(sel)
         return objs, sels
+
+    def _mixed_object_session(self):
+        """Create one molecule plus the non-molecular object kinds from #219."""
+        cmd.delete('all')
+        cmd.fragment('gly', 'mol')
+        atoms = ['mol and index %d' % i for i in range(1, 5)]
+        cmd.distance('dist01', *atoms[:2])
+        cmd.angle('ang01', *atoms[:3])
+        cmd.dihedral('dih01', *atoms[:4])
+        cmd.load_cgo([cgo.STOP], 'cgo01')
+        non_molecules = ['dist01', 'ang01', 'dih01', 'cgo01']
+        self.assertEqual(cmd.get_type('mol'), 'object:molecule')
+        for name in non_molecules:
+            self.assertNotEqual(cmd.get_type(name), 'object:molecule')
+        return non_molecules
 
     def _poll(self):
         """Run poll_panel(), returning (printed_lines, payload_from_temp_file)."""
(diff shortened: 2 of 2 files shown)


javierbq:
Thanks for this, @GhagSagar23 — closing as already fixed.

#253 landed the same guard shortly after you opened this (merged 2026-08-05), which is what closed #219 and why this now shows as conflicting. It gates on the same condition, extracted as an `is_molecule()` helper in `modules/pymol/appkit_inspector.py`.

Your diagnosis was correct and independently reached, and putting the guard in `transp_summary()` was the right call — it's the single choke point both the collapsed-row poll and the expanded-card path go through. I verified that placing it there alone does silence both, so your summary was accurate on that: `cmd.iterate` is the only probe that reaches the feedback log (it writes the selector error *without* raising), while `cmd.count_atoms` raises `CmdException` before emitting anything.

Two notes for reference:

- `cmd.get_type()` raises `CmdException` for a name that no longer exists, whereas `transp_summary()` currently never raises. Since `poll_panel()` builds `has_transp` in a single dict comprehension inside one `try`, one stale name would discard the whole payload and print an error line on every poll tick. That's why `is_molecule()` wraps the call in a `try/except` returning `False`.
- Your test covers dihedral and CGO objects, which the merged tests don't — those build only distance + angle. That coverage is worth keeping; if you'd like to open a follow-up porting `_mixed_object_session` onto `is_molecule`, it'd be welcome.

There's also a remaining piece of #219 that neither fix caught, noting it here so it isn't lost: `_build()`'s objmeta loop calls `cmd.get_title()` on every object, and for a measurement that prints `Error: object dist01 not found.` — again printing without raising, so the enclosing `try/except` can't suppress it. Expanding a measurement's card still logs that at the ~500 ms poll cadence. It needs its own issue.

Thanks again for the careful write-up.

Issue #136, opened 2026-07-08
Mutagenesis wizard has no visible UI — internal_gui setting is off


zbjones:
Environment:
  - RayMol.app 1.5.2 (16)
  - macOS (26.5.1)

  Steps to reproduce:
  1. Load any structure (e.g. fetch 6f8p)
  2. Run wizard mutagenesis
  3. Select a residue, e.g. via cmd.get_wizard().do_select("A/159/")

  Expected: The mutagenesis wizard's side panel appears, showing residue mode selection, rotamer list, and Apply/Done/Clear buttons for interactive use.

  Actual: No wizard panel is visible anywhere in the app. The wizard is fully functional under the hood — cmd.get_wizard() returns a valid pymol.wizard.mutagenesis.Mutagenesis object, do_select() and apply() work correctly via the Python API (confirmed: mutated ARG159→ALA on PDB 6f8p successfully) — but there is no GUI surface for a user to drive it without scripting.

  Root cause (suspected): cmd.get("internal_gui") returns off in this build, while internal_prompt is on. It looks like RayMol disables PyMOL's internal GUI panel (perhaps for its own custom chrome) but hasn't wired up an equivalent panel for wizards like Mutagenesis, leaving them backend-only.

  Suggested fix: Either re-enable internal_gui for wizard sessions, or implement a native RayMol-side panel that surfaces wizard.state/controls when a wizard is active.

by the way: thank you so much! Very encouraged by the work so far on RayMOL, it looks fantastic!  Would be great to have access to the mutagenesis wizard as that is a standard component of my workflow.

javierbq:
Hi @zbjones, Yes the mutagenesis wizard UI components haven't been ported yet. 

I am reimplementing all UI/UX elements in Swift to keep components portable to iOS (hopefully coming soon; I am waiting on Raymol's submission to the App Store for macOS before I submit the application for iOS). 

One of my goals for this project is to make Raymol multiplatform(as in iPhone, iPad, and Mac), and for this, it needs to be developed "Mobile First". It's harder to get UI/UX components that cram a lot of information into a small screen, but once you've done it is easy to reuse the elements in larger screens. 

Time needs to be spent  sketching UX for the tool that is intuitive and functional across platforms, if you have ideas please share them.

javierbq:
[...]

PR #252, opened 2026-08-05
Prototype a mobile-first mutagenesis wizard UI

GhagSagar23:
## Summary

This review-oriented first milestone:

- defines a platform-neutral contract for Mutagenesis wizard state, actions, command availability, and responsive layout
- adds a mock-backed SwiftUI prototype for compact and regular widths
- models residue-mode selection, rotamers, and Apply, Clear, and Done controls without invoking PyMOL
- documents the proposed interaction model and adds focused XCTest coverage
- registers the new application and test sources in the Xcode project

## Problem

RayMol's PyMOL backend supports the Mutagenesis wizard, but users cannot operate it through the native interface.

## Root cause

`PyMOLBridge.mm` disables PyMOL's `internal_gui`, while the SwiftUI application does not provide an equivalent Mutagenesis surface. The wizard therefore remains accessible only through Python commands.

## Approach

The new contract represents inactive, awaiting-residue, loading, ready, and error states and derives which commands are available in each state. Changing residue mode clears stale rotamer selection, and Apply is enabled only for a valid selected rotamer.

The prototype stacks controls below 560 points and uses a two-column layout at wider sizes. Its controller forwards enabled actions to an injected sink for testability, but intentionally performs no live bridge calls or structural mutations.

## Testing

- `git diff --check` passed.
- The configured Docker source-contract check was attempted through `scripts/safe_project.py`, but could not start because the Colima Docker socket was unavailable.
- `xcodebuild -project swiftui/PyMOLViewer.xcodeproj -scheme UnitTests_macOS -configuration Debug -destination 'platform=macOS' test` was not run because no reviewed isolated macOS runner was available and the configured Python Docker image cannot provide Xcode.

Swift compilation and the added XCTest suite therefore remain unverified.

## Risks and follow-up

- The prototype is deliberately not connected to `ContentView`, so it does not yet make the wizard available in the application.
- The final presentation location and interaction model still need maintainer approval.
- Live PyMOL state observation, command dispatch, navigation integration, and cross-platform UI validation are deferred.
- An isolated macOS runner is required to build the SwiftUI targets and execute the tests.

Relates to #136.

### DIFF
--- a/swiftui/PyMOLViewer.xcodeproj/project.pbxproj
+++ b/swiftui/PyMOLViewer.xcodeproj/project.pbxproj
@@ -12,6 +12,11 @@
 		0C0E67D9436CC6DF29DE983C /* DesignEditingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 05E6ECE945556A6D1FAE98F9 /* DesignEditingTests.swift */; };
 		0F6C6BD0DB9B077E51D1EFAF /* WhatsNewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3F5377EE69A856479317FD61 /* WhatsNewModel.swift */; };
 		0FFCBEB460CD541E932326AB /* Theme.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7D3E805AC2D63B01AB7EE29C /* Theme.swift */; };
+		A13600000000000000000011 /* MutagenesisWizardContract.swift in Sources */ = {isa = PBXBuildFile; fileRef = A13600000000000000000001 /* MutagenesisWizardContract.swift */; };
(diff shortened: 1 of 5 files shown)


javierbq:
Hi @GhagSagar23. There is a new design tool in 1.9.0 that was already released for MacOS and the UI is already ported to iOS. It will be out in the next iOS release. The design tool uses proteinmpnn, it should prediction aminoacid propensities per position and allows repacks the structure automatically after design.

The functionality of the old mutagenesis wizard I consider outdated so I don't plan to integrate it into RayMol.

I appreciate the time and effort put here; please open new feature issues so we can align on direction before implementation. 

If you haven't, please also join the Slack channel, where we can discuss new features with other members.


image

Issue #249, opened 2026-08-05
Design mode on iOS/iPadOS (#217 Phase 2d): enable RAYMOL_MPNN, ship weights, clear the MLX Simulator abort


javierbq:
Follow-up to #217, which shipped Design mode for macOS in 1.9.0 (phases 2a/2b/2c — #221, #227, #228). This ticket is Phase 2d: bringing it to iPhone/iPad.

## Why this is its own ticket

#217's v1 perf criteria were written against an iPhone 15 ("< 3 s for < 500 aa, < 10 s for < 1500 aa"), but nothing in the feature reaches iOS today. The Swift logic is already shared and compiles under a flag — the gap is packaging and an MLX runtime blocker, not the design UI.

## Current gating (all three must move together)

- `swiftui/project.yml:88` — `SWIFT_ACTIVE_COMPILATION_CONDITIONS[sdk=macosx*]: $(inherited) RAYMOL_MPNN`. iOS never defines `RAYMOL_MPNN`, so every `#if RAYMOL_MPNN` block compiles out.
- `swiftui/project.yml` dependency — `proteinmpnn-mlx` / `MPNNKit` carries `platforms: [macOS]`, keeping it out of the iOS link entirely.
- `swiftui/project.yml` copy script phase — `case "$PLATFORM_NAME" in macosx*) ;; *) exit 0 ;; esac`, so `MPNN.mpnnpack` (24 MB: `weights/design.safetensors`, `weights/packer.safetensors`, `geometry.safetensors`) is never copied into the iOS bundle.

## Known blocker: MLX aborts in the iOS Simulator

MLX aborts with `MTLStorageModePrivate is required for heaps` on iPhone 17 Pro Simulator / iOS 26 / mlx-swift 0.31.6. The documented CPU fallback does **not** prevent the abort. Verified during #217 Phase 2d exploration and recorded in #239.

Consequences:
- Inference must be validated on a **physical device**, not the Simulator.
- Any Simulator-based CI or XCUITest must not attempt to run inference (see #239, option 3).

## Scope

- [ ] Define `RAYMOL_MPNN` for the iOS SDK; drop the `platforms: [macOS]` filter once mlx-swift's iOS slice is confirmed working on device.
- [ ] Extend the `MPNN.mpnnpack` copy phase to iOS, or move the weights to an on-demand download (24 MB against the app-download budget is a real product decision — see also #224's weight-manager work, which may be the better home for this).
[...]

Issue #217, opened 2026-07-20
Interactive on-device design tool: logit-colored backbone + per-position AA editing with live repack


javierbq:
## Summary

A **stateless, interactive protein-design tool** inside RayMol, powered on-device by
[MPNNKit](https://github.com/javierbq/proteinmpnn-mlx) (LigandMPNN sequence design +
side-chain repack, already validated and benchmarked on iPhone 15 Pro). Color the backbone
by per-residue design confidence (logits), edit residues or selections from an amino-acid
palette, and on each change re-run design/repack **on-device** — every result becomes a new,
auto-aligned object so iterations can be compared directly. No server, no job queue.

## Motivation

Fast, tactile design iteration directly in the viewer where the coordinates already live.
MPNNKit already runs the full design + repack pipeline on the phone (design 154 ms @68 res →
4.3 s @2120; total 235 ms → 7.3 s on an iPhone 15 Pro), so an interactive loop is feasible.

## User-facing behavior (v1)

1. **Logit coloring.** A "color by design confidence" mode colors the backbone by the
   per-residue log-probability of its *current* amino acid under the model (low = the model
   dislikes this residue at this position). Ships with a color ramp + legend.
2. **Hover → per-position probabilities.** Hovering a residue shows a popover of the 20 amino
   acids and their probabilities at that position (sorted; current AA marked).
3. **Single-position edit.** Assign an AA to a position from a palette → the tool fixes that
   choice, re-runs **repack** (side chains) + recomputes logits, and recolors.
4. **Selection-based redesign.** Pick a named selection from a dropdown and choose target
   AA(s) from a palette → the selection is **redesigned + repacked** with the rest of the
   protein held fixed.
5. **New object per result.** Each new (repacked) sequence spawns a new RayMol object
   (auto-named, superposed onto the parent), so iterations overlay for comparison.
6. **"Predict" in the `A` menu.** Runs structure prediction on the current sequence, auto-aligns
   the predicted model to the design, and loads it as a guide object for iterating.

## Constraints / non-goals

- **Stateless.** No jobs panel, queue, or history UI — each action is a synchronous op that
  produces a new object.
- **Performance (iPhone 15).** A design/repack/re-logit operation should complete in
  **< 3 s for < 500 aa** and **< 10 s for < 1500 aa**.

## Implementation notes

Engine = MPNNKit (`MPNNModel`). The following capabilities likely need to be added/exposed in
MPNNKit to hit the UX + perf budget (track as dependencies):

[...]

PR #269, opened 2026-08-12
feat(predict): pluggable structure-prediction backend with cached model weights (#224)

### LINKED ISSUE #224: Structure-prediction backend: pluggable predictor infrastructure + cached model weights
# Structure-prediction backend: pluggable predictor infrastructure + cached model weights

## Summary

Add backend infrastructure to RayMol for running structure-prediction methods
from inside the application. The goal is a **generic, swappable predictor
abstraction** so that new prediction methods can be registered and used
interchangeably, plus a **model-weight manager** that downloads and caches each
method's weights once, on first use, verifying integrity by checksum.

**This ticket covers backend components only. No front-end / UI work is in
scope** (see *Out of scope*). All behaviour must be built **test-first (TDD)**.

## Motivation

RayMol currently has no way to run structure prediction. We want to integrate
several methods without each one wiring itself ad-hoc into the app. A single
abstraction lets us:

- add / remove / swap predictors behind one stable interface,
- ship the app without bundling multi-hundred-MB weights (download on demand),
- guarantee the bytes we run are the bytes we published (checksum),
- test the whole flow deterministically with the network mocked.

## Background / integration points

- No prediction code exists today (`grep` for `predict|boltz|alphafold` in
  `modules/` returns nothing).
- Python backend home: a new gerund-named module in `modules/pymol/`
  (e.g. `predicting.py`), consistent with `importing.py`, `computing.py`,
  `creating.py`, etc.
- There is prior art for "download a remote asset into a local cache dir": the
  `fetch` / `fetch_path` machinery in
  [`modules/pymol/importing.py`](../../modules/pymol/importing.py) (`_fetch`,
  `fetch_path` setting). The weight manager should follow the same spirit
  (configurable cache location, download-if-missing).
- Tests live in `testing/tests/<category>/`, subclass
  `pymol.testing.PyMOLTestCase`, and run via
  `pymol -ckqy testing/testing.py --run tests/<category>/<file>.py`.
- Optional: expose predictors as tools through `modules/raymol_mcp/` in a
  follow-up (not required here).

### In scope (backend)

1. **Predictor abstraction** — a common interface / base class that every
   prediction method implements, plus a registry to look one up by name and
   swap between them.
2. **Model-weight manager** — lazy, first-use download of a method's weight
   bundle (expected to be a zip on a server, URL TBD), cached once on disk,
   verified by checksum, and extracted/prepared for use. Re-runs use the cache.
3. **Configuration / registration** — how a predictor declares its identity,
   its weight bundle (URL + checksum + size), inputs, and outputs.
4. **Backend entry point / API** — a programmatic way to run a prediction by
   predictor name and get a structure back (callable from the Python API layer;
   no GUI).
5. **Error handling** — well-defined failures for network, checksum mismatch,
   unknown predictor, malformed inputs, and insufficient resources.
6. **Test suite (TDD)** — written before the implementation; deterministic,
   network-mocked.

### Out of scope

- Any UI / panel / menu / SwiftUI or AppKit work.
- Adapting any concrete/real prediction method — this ticket delivers the
  **framework only**, proven with a fake/stub predictor in the tests. Wiring up
  real predictors are follow-up tickets.
- Bundling weights in the app package.

### A. Predictor abstraction

- A base class / protocol defining the common contract shared by all
  predictors. At minimum it should express:
  - a stable **name / id** used to select the predictor,
  - a declaration of the **weight bundle** it needs (see B),
  - a **predict** operation with a well-defined input (e.g. sequence(s) /
    input spec) and output (a structure RayMol can load, e.g. a path to a
    written coordinate file or an in-memory structure object),
  - a way to report **availability / capabilities** (e.g. runs on this
    device? weights present?).
- A **registry** to register predictors and retrieve them by name, so callers
  depend only on the interface and predictors are swappable without touching
  call sites.

### B. Model-weight manager

- **Lazy download on first use**: weights are fetched the first time a
  predictor is actually run, not at install/startup.
- **Cache once**: subsequent runs reuse the cached weights with no re-download.
  Cache location configurable (a setting analogous to `fetch_path`), with a
  sensible per-user default; document the default path.
- **Checksum verification**: after download, verify the bundle against a known
  checksum (algorithm TBD — e.g. SHA-256) before use; a mismatch is a hard
  failure and must not leave a corrupt cache in place.
- **Zip bundle handling**: download the archive, verify, then extract/prepare
  into the cache; the on-disk layout after extraction is deterministic.
- **Integrity / resumption**: partial or interrupted downloads must not be
  mistaken for a valid cache (e.g. download to a temp path and atomically move
  into place only after checksum passes).
- **Idempotency / concurrency**: a second run while a download is in progress,
  or an already-complete cache, behaves correctly (no double download, no
  corruption).
- **Re-validation**: detect and recover from a cache whose checksum no longer
  matches (re-download).

### C. Configuration / registration

- A declarative description per predictor of: id, human-readable name, weight
  bundle (URL, checksum, expected size), and any runtime requirements.
- Registration mechanism that makes a predictor discoverable through the
  registry.

### D. Backend API surface

- A programmatic entry point (Python API) to: list available predictors, query
  whether a predictor's weights are cached, trigger the run (which lazily
  downloads weights if needed), and receive the predicted structure.
- No command-line-only assumptions that block later UI wiring, but **no UI in
  this ticket**.

### E. Error handling

Defined, testable failure modes for at least:

- unknown / unregistered predictor name,
- network failure during download,
- checksum mismatch,
- insufficient disk space / write failure to cache,
- malformed or unsupported prediction input,
- predictor unavailable on the current device/platform.

## Validation (TDD — required)

Follow test-driven development: write failing tests first, then implement.
Tests must be **deterministic and offline** — the network and filesystem must
be mocked/stubbed so no test hits a real server or downloads real weights.

Add a `testing/tests/predict/` category. Cover at least:

- **Registry**: register, retrieve by name, unknown-name error, swap
  implementations behind the interface.
- **Weight manager**:
  - downloads on first use, then serves from cache on subsequent use (assert
    the download path is invoked exactly once);
  - checksum success path (valid bundle accepted);
  - checksum mismatch path (rejected, cache not populated, clear error);
  - interrupted/partial download does not produce a valid cache;
  - correct extraction layout for a zip bundle;
  - re-download when a cached bundle fails re-validation.
- **Predictor contract**: a fake/stub predictor exercises the full flow
  (declare weights → lazy fetch → run → return structure) end-to-end with
  everything mocked.
- **Error handling**: each failure mode in section E raises the defined,
  documented error.
- Tests subclass `pymol.testing.PyMOLTestCase` and run under the existing
  runner (`pymol -ckqy testing/testing.py --run tests/predict/...`).

## Acceptance criteria

- [ ] Generic predictor base class / protocol + registry exist; predictors are
      swappable behind the interface.
- [ ] Weight manager downloads once on first use, caches, and reuses the cache.
- [ ] Downloaded bundles are checksum-verified; mismatches fail safely and
      leave no corrupt cache.
- [ ] Zip bundles are extracted into a deterministic, documented cache layout.
- [ ] Interrupted downloads and concurrent runs are handled without corruption
      or double download.
- [ ] All error modes in section E are defined and covered by tests.
- [ ] `testing/tests/predict/` suite is written test-first, runs offline, and
      passes under the standard runner.
- [ ] No UI/front-end changes included.
- [ ] Default cache location documented.

## Open questions / decisions to resolve during implementation

- **Weights hosting**: final URL(s) and hosting for the zip bundles (TBD).
- **Checksum algorithm** (SHA-256 assumed unless decided otherwise).
- **Cache location**: exact per-user default path and the name of the setting
  to override it.
- **Sync vs async / progress reporting**: does `predict` block, and how is
  download progress surfaced to a (future) caller?
- **Structure hand-off**: predictor returns a file path vs. an in-memory
  object loaded into the RayMol session — pick one for the interface.
- **Platform gating**: how on-device-only predictors advertise unavailability
  on unsupported platforms.
- **First predictors to wire up** in follow-up tickets — informs how much the
  interface must generalize now.

## References

- Download/cache prior art: `modules/pymol/importing.py` (`_fetch`,
  `fetch_path`).
- Testing conventions: `AGENTS.md` / `CLAUDE.md` → Testing section.

javierbq:
Closes #224.

Adds a pluggable structure-predictor abstraction plus a predictor-agnostic,
checksum-verified model-weight cache, and wires **boltz-mlx** behind it end to end on
macOS. Design: [`docs/superpowers/specs/2026-08-11-structure-prediction-backend-design.md`](docs/superpowers/specs/2026-08-11-structure-prediction-backend-design.md).
Plan: [`docs/superpowers/plans/2026-08-11-structure-prediction-backend.md`](docs/superpowers/plans/2026-08-11-structure-prediction-backend.md).

## Verified end to end

```
predict boltz2, MKTAYIAKQRQISFVKSHFSRQLEERLGLIEVQ, name=trp2
```

529 MB fetched from the release → checksum verified → three members extracted → `.ok`
sentinel written → Swift/MLX inference → **276 heavy atoms, 33 residues, chain A**, first
residues `MET LYS THR ALA TYR` matching the input, loaded into the session. A second
submission served the weights from cache with no re-download.

- Python: **69 tests** in `testing/tests/predict`
- Swift: **201 tests**, 0 failures (macOS) — and the iOS slice compiles
- Both slices hand-compiled, because **no CI job builds Swift**

## The architecture, and why

**RayMol has no Python→Swift call path.** `grep -rn "@_cdecl\|@convention(c)" swiftui/`
returns zero hits, `PyMOLBridge.h` is one-directional by construction, and Design mode is
UI→Swift→Python only. Rather than build that missing direction (a new C ABI plus GIL work
under `_PYMOL_EMBEDDED`, where `PAutoBlock` is a no-op), the job-handle API makes it
unnecessary: `cmd.predict` writes a request JSON and prints a `PREDICT:` marker that
`pollFeedback()` already scans on its 100 ms timer — the same mechanism `OBJPANEL:` and
`SETTINGS:ready` use. Swift writes status/result files that Python polls.

Consequence, deliberate: `cmd.predict` only *runs* inside the app. Headless `pymol -c`
reports the predictor unavailable, which is what satisfies #224's "unavailable on this
platform" failure mode and what makes the offline suite possible.

## Notable

- **Inference defaults are upstream Boltz's** (recycling 3 / 200 steps), not the MLX port's
  (0/20), which **fail the port's own quality gate** at 3.19 Å / 0.685 lDDT. `step_scale`
  needs no work — it comes from the artifact's `config.json`, already 1.5.
- **`diffusion_samples` is rejected, not ignored.** The port doesn't plumb it and only
  sample 0 escapes; silently accepting a quality knob produces results the user believes
  are something they aren't.
- **`PredictSizeGuard` is preventive**, mirroring `DesignSizeGuard`: boltz's own preflight
  under-predicts measured peaks by 10–25× and can't fire at all under its phone defaults,
  while jetsam is an uncatchable SIGKILL that on macOS costs the user their session.
- **`MLXRuntime`** (this branch) arbitrates the process-global MLX cache limit min-wins, so
  boltz's `MemoryPlanner.apply()` — which assigns it on *every* predict — can't raise Design
  mode's 96 MB ceiling by call order.
- **Upstream**: `javierbq/boltz-mlx` gained a public `StructureWriter` (`5aec62e`) and its
  first tag `v0.1.0`; weights published as release `weights-v1`.

## Bugs this work found

1. **`colorprinting.info` doesn't exist.** All six message sites used it, and `quiet=0` is
   the *default command-line path*, so every new command crashed after submitting. The suite
   was 48/48 green because it only exercised `quiet=1`. Fixed, plus a reflective guard that
   asserts every `colorprinting` helper `predicting.py` reaches for actually exists.
2. **A C `setenv` after `Py_InitializeFromConfig` is invisible to `os.environ`.** `C getenv`
   returned `b'1'` while `os.environ` returned `None`, so the predictor reported itself
   permanently unavailable in-app.
3. **`%2@` does not pad** in Foundation's `String(format:)` — every PDB record came out 77
   columns instead of 78. (MPNNKit's shipped `PDBWriter` uses the same `%2@`.)
4. **Peak memory is *sub*-linear in tokens** over the measured range; a quadratic fit sat
   *below* measurement, which is exactly what licenses a run that dies.

## Reviewer notes

- **Timing is a Debug measurement**: ~95 s cold / 73.5 s warm for 33 residues at 200 steps,
  `-Onone`. An upper bound, not comparable to boltz-mlx's Release figures, and worse than
  extrapolating the 50-step numbers — for short sequences the fixed per-step cost of 200
  steps dominates the ~N² term. Re-measure against Release before sizing any UI.
- **In-process inference is an accepted residual risk.** `MLXRuntime.withMLXErrorsAsThrows`
  makes MLX-reported errors catchable, but jetsam is not — so an OOM the guard doesn't catch
  kills the app and the user's unsaved session. Out-of-process isolation (for boltz *and*
  MPNNKit) is the follow-up.
- **iOS is intentionally not enabled**: MLX cannot run in the iOS Simulator at all, and no
  boltz run above ~115 tokens has ever completed on a physical device because iOS suspends
  the app mid-run.
- CI now runs the `testing/tests/predict` **directory** rather than individual files — the
  only variant that can't re-create the enumeration gotcha that PR #259 had to repair.

Deferred to follow-ups: iOS enablement, out-of-process isolation, MSA support, migrating
#249's `MPNN.mpnnpack` onto `WeightCache`, and mmCIF output.


javierbq:
Follow-ups opened for the items this PR defers:

- #273 — run MLX inference out-of-process, so an OOM fails the job instead of the session. Covers MPNNKit too, which carries the same exposure today.
- #274 — real MSA support (blocked first on resolving a contradiction in boltz-mlx: its prose says MSAs are refused, its code implements them with bitwise parity tests).
- #275 — move `MPNN.mpnnpack` onto `WeightCache`, closing the loop on the predictor-agnostic requirement from #224's triage.
- #276 — emit mmCIF, lifting PDB's 26-chain and column limits.

iOS enablement is deliberately **not** filed — not planned.

javierbq (commented):
## Code review — pluggable structure-prediction backend (#224)

**Scope reviewed:** ~1,900 lines of code (Python predictor framework + weight cache; Swift job manager, MLX runtime arbitration, size guard) plus tests, build config, and docs. Method: 8 area-reviewers → 8 adversarial verifiers against the *real* code (including the resolved boltz-mlx `v0.1.0` sources), plus two facts confirmed empirically.

**Verdict:** Architecturally sound and the security-critical core is solid. The earlier machine-local-path merge-blocker is **already fixed** on this branch. One functional gap (cancel) and one contract/UX mismatch (sync download) are worth fixing before merge or as fast follow-ups; the rest are low-severity robustness/observability/test-coverage items.

---

### What's solid (verified, not assumed)

- **Weight-cache integrity core** — sha256 is verified on the complete `.part` *before* extract/publish; the `.ok` sentinel is content-checked (not existence-checked), so truncated/corrupt caches are correctly rejected and re-downloaded; `os.replace` + sentinel-last publish leaves no window where `is_cached()` accepts a partial dir; zip-slip is closed by the exact-member-set assertion. Error taxonomy (network vs unwritable-cache vs bad-zip) is correct.
- **Merge-blocker resolved** — `project.yml` → `BoltzMLX url: …/boltz-mlx.git, from: 0.1.0` (no machine-local `path:`); `project.pbxproj` uses an `XCRemoteSwiftPackageReference` to that URL; `Package.resolved` pins `0.1.0`/`5aec62e`.
- **Wire contract** host.py ↔ `BoltzJobManager.Request/Status/Chain` key names match; `RAYMOL_PREDICT_HOST` is set **before** `Py_InitializeFromConfig` (the setenv trap is handled); the iOS slice compiles clean (all Boltz symbols `#if os(macOS)`); CI's `testing/tests/predict` directory entry expands correctly.
- **Tests avoid the known traps** — explicit `quiet=0` tests + a reflective guard that asserts every `colorprinting` helper used actually exists; real-thread concurrency test with a barrier (not mocks).

---

### Findings (deduplicated, by severity)

#### 🔴 High

**1. Cancel never interrupts a running inference** — `swiftui/PyMOLViewer/Shared/BoltzJobManager.swift:218`
`awaitSync` runs `predictor.predict(...)` inside a detached `Task {}` whose handle is discarded and **never `.cancel()`d**. boltz-mlx's `predict()` guards each diffusion step with `try Task.checkCancellation()` — but that reads the *running* task's flag, which nothing ever sets. The `cancelled` Set is only consulted at coarse phase boundaries (`:170`, `:198`), and `:198` runs *after* `predict()` fully returns.
- **Effect:** a cancel during inference does **not** stop compute — the run continues to completion (seconds→minutes, holding multi-GB MLX peak, potentially triggering the very jetsam the size guard exists to prevent), and the finished result is merely discarded afterward. `host.py:71`'s docstring ("lands on the per-diffusion-step `checkCancellation`, so worst case is one step") is false as wired.
- **Fix:** keep `runningTasks: [String: Task<…>]` keyed by jobID under `stateQueue`; on a cancel marker `.cancel()` the live task. Then the existing `catch is CancellationError` (`:204`) reports "cancelled" promptly. Correct the host.py docstring too.

#### 🟠 Medium

**2. `predict()` docstring says "returns immediately" but the first call blocks the main thread on a synchronous ~529 MB download** — `modules/pymol/predicting.py:39` (`:105` runs before `submit()`)
On a cold cache, `weight_cache().ensure()` streams + extracts the whole bundle *in the calling thread* before `submit()`. `BoltzJobManager`'s own comment notes `cmd.predict` from the console arrives **on the main thread**, so first use beachballs the window and the 100 ms `pollFeedback` timer — and with the default `quiet=1` the progress callback is gated off, so it's silent. The docstring's own EXAMPLE triggers it.
- **Fix:** either move `ensure()` off the render thread (submit first; fetch weights in the background and surface a "downloading" status), **or** amend the docstring/NOTES to say the first uncached call blocks on a ~505 MiB download and point users to `predict_weights(predictor, download=1)` to pre-warm.

**3. `requireCacheLimit` installs `MLX.Memory.cacheLimit` *outside* the lock — lost-update race defeats min-wins** — `swiftui/PyMOLViewer/Shared/MLXRuntime.swift:107`
`effective` is computed under `NSLock` (`:96`) but the global assignment is at `:107`, after `unlock()`. With Design (`MPNN`/96 MB) and Predict (`Boltz`/256 MB) both live on macOS and running on different threads, interleaving can leave `cacheLimit = 256` installed while the registry min is 96 — exactly the "a later larger request raised a ceiling someone needs low" regression this type exists to forbid. Serial-only `MLXRuntimeTests` can't catch it.
- **Fix:** trivial — assign `MLX.Memory.cacheLimit = effective` inside the locked region.

#### 🟡 Low (real, narrow)

**4. Out-of-range integer options silently hang a job "queued" forever** — `modules/pymol/predictors/base.py:47` + `BoltzJobManager.swift:111`
`seed`/`recycling_steps`/`diffusion_steps` are only lower-bounded in Python (unbounded ints), but Swift decodes `seed` as `UInt64` and steps as `Int`. e.g. `diffusion_steps=10**30` passes Python, overflows `JSONDecoder`, and `handle()`'s `guard let … = try? …` returns **without writing a status** — asymmetric with the preflight path which *does* write "failed". `predict_status` then reports `queued` forever. The same `try?` silently eats any future wire-format drift too.
- **Fix:** write a `failed` status when `parseRequest` throws (closes the whole class of silent deaths), and/or add matching upper bounds in `PredictionOptions`.

**5. Weight-cache lock coordinator: no ownership/liveness check** — `modules/pymol/predictors/weights.py:160`
Two related issues: (a) the pid written into the lock (`:150`) is **never read back** for liveness, and `LOCK_TIMEOUT`(900) < `LOCK_STALE_SECONDS`(1800), so the first caller arriving within 900 s of a crashed downloader spins 15 min then fails with a misleading "another download in progress." (b) A live download exceeding 1800 s (unbounded — the 30 s timeout is per-read) can be wrongly reclaimed by a second caller; since temp files are keyed only on `sha256`, they collide and the `finally` cross-deletes the other's files. *Integrity is preserved* (each process hashes its own stream), so harm is spurious failures + wasted re-downloads, not corrupt weights.
- **Fix:** `os.kill(pid, 0)` liveness check in the `FileExistsError` branch; set `LOCK_STALE_SECONDS ≤ LOCK_TIMEOUT`; key temp files on pid/token; verify pid before `_release_lock`.

**6. `MemoryPlanner.apply()` pins `Memory.memoryLimit=6 GB` — never arbitrated, never restored** — `BoltzJobManager.swift:196`
boltz-mlx sets both `cacheLimit` *and* `memoryLimit` on every predict; RayMol only arbitrates/re-asserts `cacheLimit`. So `memoryLimit` stays at 6 GB for the rest of the process. On a **thrown** predict the `:196` re-assert is skipped, leaving both stale until the next `configureOnce` (cacheLimit) / forever (memoryLimit).
- **Fix:** re-assert in the `catch` block (`:206`); bring `memoryLimit` into arbitration or document that RayMol accepts the 6 GB pin.

**7. `RAYMOL_WEIGHTS_DIR` documented as "published by the Swift host" — but nothing sets it** — `modules/pymol/predictors/weights.py:85` (+ design spec)
A full-tree grep finds zero native writers; the only env the bridge publishes is `RAYMOL_PREDICT_HOST`. The cache always falls through to `default_root()` (which lands in the sandbox container only via `expanduser('~')`). Misleading mental model.
- **Fix:** either `setenv("RAYMOL_WEIGHTS_DIR", …)` in `PyMOLBridge.mm`, or correct the docstring/spec.

**8. `cancelled` Set grows unbounded** — `BoltzJobManager.swift:107`
Cancel markers for jobs that failed preflight (never entered `run()`) or arrived post-completion are inserted and never removed. Tiny but monotonic. Fix: gate insert on a live job / evict on terminal status.

#### Test-coverage gaps (all confirmed, low)

- **Cancellation effect is untested** (`BoltzJobManagerTests.swift:154`) — only `parseMarker` is asserted; `handle(.cancel)` routing and the `run()` flag observation are unexercised, and `cancelled` is private with no seam. A regression could no-op cancel with every test green. (Pairs with finding #1.)
- **Disk-full errno → `WeightCacheUnwritable` sub-branch untested** (`weights.py:231`, `:262`) — the `ENOSPC/EACCES/EROFS` classification during download/extract is never hit.
- **Malformed/partial status-file JSON fallback untested** (`host.py:64`) — the `ValueError` arm guarding a mid-write read on the 100 ms poll is unexercised; narrowing the `except` would leak a traceback out of `pollFeedback`.
- **`predict_weights(download=1)` prefetch path untested** (`predicting.py:214`).

#### Nits / docs

- **`warn` tier is behaviorally dead in the predict path** (`BoltzJobManager.swift:125`) — `preflight` maps both `.ok` and `.warn` to "proceed," so `okFraction=0.50` has no effect and the `.warn` payload is built and discarded. (The Design path *does* surface `warn` to the user — inconsistent.) Either surface a caution or collapse to a single refuse threshold.
- **`weights.py:7,33` say "~533 MiB"** but the declared bundle is 505 MiB / 529 MB — drift, twice. Reference `bundle.size` or fix the number.
- **`host.py:71` "worst case is one step"** also understates the trunk/featurize window (only the diffusion phase is per-step checked; the trunk has zero `checkCancellation`). Fold into #1's docstring fix.

---

### Checked and cleared (looked at, not defects)

- **First status write hanging the job** — refuted: `FileManager.replaceItemAt` **creates** the target when the original is missing (tested live), matching the reported end-to-end success.
- **Size guard budgeting against total RAM instead of free RAM** — refuted: `os_proc_available_memory()` returns **0** on macOS for an unrestricted process, so the "fix" would refuse *every* predict; the 50%/75% fractions are the deliberate, documented headroom, and the fit sits at/above all 17 measured points.
- **zip-slip**, **the 1024 (Python) vs 900 (Swift) ceiling causing a hang** (preflight always writes a `failed` status), and **"default cache path not documented"** (surfaced at runtime by `predict_weights` + in the spec) — all refuted.

---

**Bottom line:** merge-ready in structure. I'd want **#1 (cancel)** and **#2 (sync download / docstring)** addressed, and **#3** and **#4** are cheap high-value hardening. Everything else is fine as fast follow-ups.

<sub>🤖 Review generated with [Claude Code](https://claude.com/claude-code) — 8 area-reviewers + 8 adversarial verifiers against the resolved boltz-mlx v0.1.0 sources.</sub>

javierbq:
Addressed in 880924804. Verified each finding against the code first; all eight were real.

**#1 cancel (high)** — confirmed: `awaitSync` discarded the `Task` handle, so boltz-mlx's per-step `Task.checkCancellation()` read a flag nothing ever set. Tasks are now registered per job id under `stateQueue` and a cancel marker cancels the live one. **Mutation-checked**: reinstating the discard fails the new test with an async-wait timeout, so it bites rather than passing vacuously. `host.py`'s docstring corrected too — and it also understated the trunk window, which has no cancellation points at all, so that is now stated.

**#2 sync download (medium)** — docstring corrected and it now points at `predict_weights(download=1)`. Also emitting a warning *regardless of* `quiet`, since the real defect was silence during a multi-minute main-thread stall, not just the wrong docstring.

I did **not** move `ensure()` off-thread. A background *Python* thread is GIL-starved roughly 1000x in the embedded app — measured during the benchmark, where a zip extraction ran at ~364 KB/s because the main thread holds the GIL persistently. That option is worse than the disease; doing it properly means a Swift-side downloader, which contradicts "Python owns the weight manager." Flagging in case you want that tradeoff revisited.

**#3 cacheLimit race (medium)** — install moved under the same lock as the min computation.

**#4 silent death (low)** — an unparseable request now writes a `failed` status at the conventional path, closing the class rather than the instance. Options are additionally bounded at the top, since the concrete trigger was overflowing Swift's `Int`/`UInt64` decode.

**#5–#8** — pid liveness via `os.kill(pid, 0)` so a crashed downloader is reclaimed immediately instead of after a 15-minute spin; `LOCK_STALE_SECONDS <= LOCK_TIMEOUT`; pid-keyed scratch paths so two processes cannot cross-delete; `cacheLimit` re-assert moved into a `defer` so a *thrown* predict no longer leaves the larger ceiling installed (the 6 GB `memoryLimit` pin is now documented as accepted); `RAYMOL_WEIGHTS_DIR` docs corrected — nothing sets it, `default_root()` is what runs; cancel records no longer accumulate for terminal jobs; the "~533 MiB" drift now references `WeightBundle.size`.

**Test gaps** — all four closed, plus the new bounds: +6 Python, +5 Swift.

**One I answered rather than implemented.** The dead `.warn` tier is now documented, not revived or collapsed. Adding a caution field to the wire format for a feature with no UI is speculative; the shape is shared with `DesignSizeGuard`, whose controller does surface a caution, and #217's deferred Predict menu item is where it belongs. Happy to collapse to a single threshold instead if you'd rather not carry an unused tier.

macOS 212 tests / 0 failures · iOS `BUILD SUCCEEDED` · Python 75 tests. Both slices hand-compiled, since no CI job builds Swift.

Also noting your four refutations were right and I did not "fix" any of them — particularly `os_proc_available_memory()` returning 0 for an unrestricted macOS process, which would have made the size guard refuse every predict.

Issue #297, opened 2026-08-15
Feed an MSA to a predictor: a capability on the base class, not a boltz2 special case


javierbq:
## Summary

Wire the MSA objects into `cmd.predict` and on to
`BoltzFeaturizer.featurize(_:alignments:)` — through a **declared capability on the
`Predictor` base class**, so a method that cannot use an alignment refuses by name instead of
quietly folding single-sequence.

Second of three (#274). Depends on #296 for the noun.

## The contradiction that blocked this is resolved

**The code was authoritative; the prose was stale.** boltz-mlx implements the whole MSA path:
`featurize(_:alignments:)` takes an optional per-chain map and falls back to
`MSAAlignment.singleSequence` for a chain that has none, `MSAAlignment.swift` is a 273-line
port of upstream's `_parse_a3m` / `construct_paired_msa` / `_prepare_msa_arrays_inner`,
`msaLengthMismatch` and `msaQueryMismatch` exist and are thrown, `BoltzInputLimits.desktop`
admits 16384 rows (upstream's `const.max_msa_seqs`), and 39 tests across four files exercise
it. Fixed upstream in javierbq/boltz-mlx#1, merged.

Two corrections to the record while we are here:

- **The README never claimed MSAs were out of scope.** Only the `BoltzFeaturizer.swift` header
  did. Its "out of scope" line covers confidence, affinity, potentials and training.
- **"`MSAFeaturizerParityTests` assert bitwise parity" is weaker than it reads.** Those tests
  and `MSAEndToEndTests` `XCTSkip` when the reference bundles and the barnase/barstar a3m under
  `.artifacts/` are absent; nothing in that repo rebuilds them, they are not committed, and
  boltz-mlx has no CI workflows at all. A green `swift test` there is not evidence of parity.
  Do not cite it as acceptance for this ticket — verify on the RayMol side.

## Design — the pluggable part

One class attribute and one hook on `Predictor` (`modules/pymol/predictors/base.py:99`):

```python
class Predictor(abc.ABC):
    #: True if this method can genuinely use a real MSA. Default False, so a predictor
    #: that cannot is rejected BY NAME. Accepting an alignment and folding
    #: single-sequence anyway produces a worse structure with nothing saying so --
    #: the same failure mode validate_options() already exists to prevent for knobs.
    supports_msa = False

    def bind_alignments(self, spec, alignments):
        """Attach {chain_id: MSAObject} to `spec`, or raise PredictionInputError.

[...]

Issue #296, opened 2026-08-15
MSA objects: load an alignment once, reuse it across targets and methods


javierbq:
## Summary

Make a multiple-sequence alignment a **named, session-stored thing in its own right**, the
way a structure is, instead of an argument that evaporates after one call. `load_msa` reads
an a3m (or aligned FASTA) from disk into an MSA object; the object survives a `.pse` round
trip, appears in the object panel, and can be attached to a target object and chain.

First of three (#274). Nothing here consumes an MSA (#297) and nothing here generates one
(#298). This ticket is the noun both of those need.

## Motivation

An alignment is expensive — minutes of MMseqs2 search, or a file someone curated by hand —
and it is a property of a *sequence*, not of a run. Passing an a3m path per `predict` call
means it is re-read and re-validated every time, and it is invisible afterwards: nothing in
the session records which alignment produced which model, and folding one target with both
`boltz2` and `boltz2-bf16` supplies it twice with no guarantee the two are the same bytes.
Naming it fixes all three, and is the difference between "MSA support" and an MSA the user
can actually work with.

## Design

**Python-level and session-persisted, following `raymol_notes.py` — not a C++ `CObject`.**
An MSA has no geometry, so every one of `CObject`'s virtuals (extent, matrix, render, state
count) would be a stub, and an entry in the object list that accepts `show cartoon` and does
nothing is a worse lie than a section that plainly isn't a molecule. What that buys —
`delete` / `set_name` / groups for free — is three small commands' worth of work. The
precedent is well worn here: `raymol_notes`, `raymol_scenes` and `raymol_scene_anim` all
carry non-geometric state through a `.pse` this way, and the object panel already merges
Python-side state (`predicting._PENDING`) into its rows.

New files, mirroring the shape of `predictors/`:

| file | responsibility |
|---|---|
| `modules/pymol/msa.py` | the `cmd.*` surface. Thin, like `predicting.py` |
| `modules/pymol/msas/store.py` | the named store, plus session save/restore |
| `modules/pymol/msas/parse.py` | a3m / aligned-FASTA reader — validation and summary only |
| `modules/pymol/msas/errors.py` | `MSAError`, `MSAInputError`, `MSANotFound` |

**Every function in `msa.py` must end its signature with `_self=cmd`.** `pymol2/cmd2.py`
binds `_self` only when it appears in the argspec and otherwise copies the function verbatim,
so an omission silently drives the *global* instance — see the module docstring of
[...]

Issue #284, opened 2026-08-14
Weight download freezes the app with no UI: first predict blocks the main thread for 529 MB, so its own progress messages can't render


javierbq:
## Summary

The first `predict boltz2, …` on a machine with no cached weights freezes RayMol for the
entire 529 MB download + extraction — **with nothing on screen at all**: no message, no
spinner, no progress, no cancel. The window stops redrawing and looks crashed. On a slow
link that is minutes.

I want a sheet/popover on first use: bundle name and size, a **determinate progress bar**
(download %, then extract %), and a **Cancel** button — with the app staying responsive
throughout.

## Reproduce

```bash
rm -rf ~/Library/Application\ Support/RayMol/weights/boltz2-mlx-int8
```

Launch RayMol, type into the console:

```
predict boltz2, MVTPEGNVSLVDESLLVGVTDEDRAVRSAHQFYERLIGLWAPAVMEAAHELGVFAALAEAP
```

The app hangs. When it finally returns, every message it *should* have shown during the
wait flushes into the log at once.

## Why nothing appears

This is not a missing message — the messages are already written, and they are *structurally
unable to render*. Four things line up:

1. **`predict` runs on the main thread.** `heavyLabel(for:)`
   (`swiftui/PyMOLViewer/Shared/PyMOLEngine.swift:893-902`) matches only ray / surface /
   mesh / dots, so `runCommand` takes the synchronous branch and calls `runCommandCore`
   directly on the main thread (`PyMOLEngine.swift:843-847`).

2. **The download happens inline on that thread.** `cmd.predict` calls
   `cache.ensure(bundle, progress=report)` (`modules/pymol/predicting.py:371`), which
   streams 529,338,573 bytes in 1 MiB chunks and then unzips them
   (`modules/pymol/predictors/weights.py:227-285`).

3. **The progress messages go into a buffer nobody is draining.** `predicting.py:361-365`
   emits a pre-flight warning (deliberately ignoring `quiet`, precisely so this wouldn't
   look like a hang) and `predicting.py:367-370` emits a `predict: download NN%` line per
   chunk. Both land in PyMOL's feedback buffer. The app **pulls** that buffer from
   `pollFeedback()` via `PyMOLBridge_GetFeedback` (`PyMOLEngine.swift:2952-2961`), driven by
   a `Timer.scheduledTimer` on the main runloop (`PyMOLEngine.swift:635-644`). A blocked
   main thread cannot service its own runloop, so the timer never fires and the buffer is
[...]

Issue #274, opened 2026-08-12
Real MSAs in cmd.predict (tracking)


javierbq:
## Summary

Let `cmd.predict` fold a target with a real multiple-sequence alignment, instead of the
single-sequence mode #224 shipped — and make the alignment itself a **reusable object**, not a
per-call argument.

Tracking issue. The work is split into three, below.

## Motivation

Single-sequence folding is materially worse for anything that needs co-evolutionary signal.
The Swift side already implements the whole MSA path, so this is plumbing plus one design
decision: an alignment costs minutes of search or a hand-curated file, and it belongs to a
*sequence*, not to a run. Loading one per `predict` call would leave it invisible in the
session, re-validated every time, and impossible to reuse across methods or targets. So it
becomes a named object first, and everything else hangs off that.

## The contradiction is resolved

boltz-mlx said both that real MSAs were "out of scope and REFUSED" and, in code, implemented
them. **The code was authoritative; the prose was stale**, corrected in
javierbq/boltz-mlx#1 (merged). Two corrections to this issue's original framing:

- **the README never claimed MSAs were out of scope** — only the `BoltzFeaturizer.swift`
  header did, so the upstream fix was a one-file comment edit;
- **"bitwise parity is asserted" is weaker than it reads.** `MSAFeaturizerParityTests` and
  `MSAEndToEndTests` `XCTSkip` when the reference bundles and barnase/barstar a3m under
  `.artifacts/` are missing; none of it is committed, nothing rebuilds it, and boltz-mlx has no
  CI. A green `swift test` there proves nothing — verify on the RayMol side instead.

## The three tickets

1. **#296 — MSA objects.** A session-stored, named alignment with `load_msa`, attachable to a
   target object and chain, visible in the object panel. The noun the other two need.
2. **#297 — predictors consume MSAs.** A `supports_msa` capability on the `Predictor` base
   class plus a `bind_alignments` hook, `msa=` on `cmd.predict`, the a3m wire through `host.py`
   to `BoltzFeaturizer.featurize(_:alignments:)`, and an MSA-depth term in `PredictSizeGuard`.
3. **#298 — `msa_search`.** Build an alignment with ColabFold's MSA server, defaulting to the
   public one and configurable to a private deployment.

They are ordered by dependency, but #298 only needs #296, so #297 and #298 can run in parallel
once #296 lands.

## Non-negotiables, across all three

These are deliberate fidelity choices in boltz-mlx; "improving" any of them silently corrupts
results:

[...]

PR #300, opened 2026-08-16
msa_search: build an MSA with ColabFold's MSA server, public or private (#298)

### LINKED ISSUE #298: msa_search: build an MSA with ColabFold's MSA server, public or private
## Summary

`msa_search` runs an MMseqs2 search against a **ColabFold MSA server** and lands the result as
an MSA object. Defaults to the public `https://api.colabfold.com`; points at a private
deployment with one setting.

Third of three (#274). Depends on #296 for where the result goes; independent of #297.

## Motivation

With the first two tickets, RayMol can load an alignment and fold with it — but it cannot
*produce* one. Requiring users to run ColabFold somewhere else and copy an a3m back makes the
MSA path a power-user feature, which defeats the point of making the MSA a first-class object.

## Privacy — read this first

`api.colabfold.com` is a **third-party service**. Searching a sequence there publishes that
sequence to someone else's infrastructure, and RayMol's users fold unpublished designed
binders. So:

- the server actually used is named on every `msa_search` line and recorded in the MSA object's
  `source` provenance, so it is still visible inside a saved `.pse`;
- the first use of a **public** server in a session warns, regardless of `quiet`, that the
  sequence leaves this machine;
- the private path is first-class, not a workaround.

We already run such a deployment. `deploy_colbfold_local` provisions a GCE VM running
ColabFold's own `msa-server` binary as a systemd unit (`startup-script.sh:22-40`,
`msa-server -local -config config.json`), which means the protocol is byte-identical to the
public one and only the base URL changes. That is the whole reason to target this API rather
than shelling out to a `colabfold_search` install.

## Design

`modules/pymol/msas/colabfold.py` (protocol) and `modules/pymol/msas/searching.py`
(the background half), with `cmd.msa_search` in `msa.py`.

### Configuration

```
msa_server https://msa.internal.example    # set; persists via ~/.raymolrc.py
msa_server                                 # print the current one
```

Resolution order: explicit `server=` argument → `msa_server` setting → `RAYMOL_MSA_SERVER`
environment variable → the public default. The env var because that is how every other RayMol
backend knob works (`RAYMOL_WEIGHTS_DIR`, `RAYMOL_PREDICT_HOST`), and `~/.raymolrc.py` because
the native apps boot Python directly and never read `~/.pymolrc` at all (`raymolrc.py`, #225).

### It must not block

Same rule as the weight fetch, and for the same sharp reason. A search is minutes;
`cmd.msa_search` is reachable from the console, which runs on the main thread; and RayMol
drains PyMOL's feedback buffer from a **main-run-loop timer**, so a blocked main thread cannot
deliver even the messages describing why it is blocked. That is exactly how #284 happened.

So `msas/searching.py` mirrors `predictors/fetching.py`: the worker thread touches the
filesystem and `print`, and **nothing else** — never the session, because RayMol's Metal
renderer reads object state on the main thread without taking PyMOL's API lock. The MSA object
is created by a main-thread pump when the search lands, the way `predicting.pump()` submits a
job whose weights have arrived. `poll_panel()` already calls that pump every 500 ms
(`appkit_inspector.py:479`); this one hangs off the same tick.

An `MSA:` feedback marker alongside `PREDICT:`, `WEIGHTS:` and `OBJPANEL:`, carrying scalars
only so it stays under the ~1 KB `OrthoLineLength` cap. Plus `msa_status` and `msa_cancel`,
matching `predict_status` / `predict_cancel`, both of which also pump.

### Protocol — confirm against the deployment before coding

From ColabFold's `run_mmseqs2`: `POST {host}/ticket/msa` with the query FASTA and a mode; poll
`GET {host}/ticket/{id}` until the state is `COMPLETE`; then `GET {host}/result/download/{id}`,
a tar.gz containing `uniref.a3m` and the environmental a3m. **Verify this against
`/opt/ColabFold/MsaServer` on the deployed VM first** — the sketch is from ColabFold's client,
not from the server we run, and the two can drift by version.

Handle `RATELIMIT`, `MAINTENANCE` and `ERROR` states explicitly. A rate-limited public server
is the *common* case, and it has to read as "the server refused, try later or point at your
own", never as a hang. Send a real `User-Agent`; the public server asks for one and throttles
anonymous traffic.

### Caching

Key on sha256 of the normalised query, plus server and mode, under the same cache-root
convention `WeightCache` uses. Re-searching the same target reuses the result; `refresh=1`
forces a new search. A search that costs minutes and gets thrown away on the next call is the
thing users notice first.

## Non-negotiable

**Do not request or fabricate paired alignments.** ColabFold can return paired MSAs for
complexes, but boltz-mlx's pairing reads taxonomy only from `>UniRef100_*` headers and only
when a taxonomy DB is passed, so it is inert for anything this ticket can deliver. Marking
target-homolog rows as paired to compensate asserts co-evolution across the very interface an
interface score measures — and it fails by reading HIGH, not by crashing. Per-chain unpaired
MSAs only; real pairing is a separate ticket, with a taxonomy DB.

## Tests

`testing/tests/msa/msa_search.py`, with the network mocked at the same `_urlopen` seam the
weight tests already patch (`pymol.predictors.weights._urlopen`) — **never reach a real
server**, public or private.

- ticket → poll → download happy path, landing an MSA object with the right depth and provenance;
- `RATELIMIT` and `ERROR` states surface as errors naming the server, not as hangs;
- cancel mid-poll leaves no half-built object;
- the cache hits on a repeat search and `refresh=1` misses;
- **the worker thread creates no session object** — assert the object appears only after the pump;
- the public-server warning fires once and names the host;
- `quiet=0` as well as `quiet=1`, per `docs/predictors.md`.

## Out of scope

- Structural templates (`pdb70` hits) — a different feature with a different consumer.
- Paired MSAs (see above).
- Running MMseqs2 locally on-device.

javierbq:
Closes #298. Third of #274, on top of #296's alignment objects. Independent of #297.

## What lands

`msa_search` runs an MMseqs2 search on a ColabFold MSA server and stores the result as a named alignment object, so RayMol can now *produce* an alignment rather than only load one.

```
msa_server https://msa.internal.example   # or RAYMOL_MSA_SERVER, or the public default
msa_search 1ubq, target=1ubq              # returns at once; lands minutes later
msa_status / msa_cancel
```

New: `msas/colabfold.py` (protocol), `msas/searching.py` (background half + cache), `cmd.msa_search` / `msa_server` / `msa_status` / `msa_cancel` in `msa.py`, registered in `keywords.py`, `api.py` and the completer.

## Protocol — checked against the deployment, not just the client

`deploy_colbfold_local` runs ColabFold's own `msa-server -local`, which is MMseqs2-App's Go backend. Two states ColabFold's client treats as job states are not, and reading `status` off a 200 misses both:

- **RATELIMIT** is the answer to the *request*: tollbooth replies HTTP 429 with `{"status":"RATELIMIT","reason":...}` (`backend/server.go`), so it arrives as an `HTTPError`. On the public server this is the common case, and it now reads as "the server refused — try later or point at your own", never as a hang.
- **MAINTENANCE** comes from the public front end rather than the backend, so it is accepted from either.

`ERROR`/`UNKNOWN` and an unrecognised status are handled by name too, and every message names the host that refused. A real `User-Agent` is sent.

## Privacy

The default server is a third party. So: the server is named on every `msa_search` line and recorded in the alignment's `source` provenance (visible inside a saved `.pse`), and the first use of a public server in a session warns that the sequence leaves this machine — **regardless of `quiet`**, since `quiet=1` is the API default a scripted loop takes. The private path is first-class: only the base URL changes.

## It does not block

A search is minutes and the console runs on the main thread — exactly how #284 froze the app. `msas/searching.py` mirrors `predictors/fetching.py`: the worker touches only the filesystem and `print`, never the session, and the alignment is put in the store by a main-thread pump hanging off `poll_panel`'s existing 500 ms tick. Plus an `MSA:` feedback marker carrying scalars only (clipped so it cannot overflow the ~1 KB feedback line), and `msa_status` / `msa_cancel`, which also pump.

Results are cached on disk, keyed on sha256 of the normalised query + server + mode under the same root convention `WeightCache` uses; `refresh=1` forces a new search.

## Non-negotiable: no paired alignments

Pairing modes are refused by name, `/ticket/pair` is never called, and a `pair.a3m` in the result is ignored. boltz-mlx reads taxonomy only from `>UniRef100_*` headers and only with a taxonomy DB, so faking pairing fails by reading HIGH on interface scores, not by crashing. A multi-chain query is refused with a message telling the user to search each chain separately.

## Tests

`testing/tests/msa/msa_search.py` — 46 tests, network mocked at the `_urlopen` seam, **never** reaching a real server. Covers the ticket→poll→download happy path and provenance; RATELIMIT/ERROR/MAINTENANCE/unreachable surfacing as errors that name the server; cancel mid-poll leaving no half-built alignment; cache hit / `refresh=1` miss / different server or mode as different entries; **the worker creating nothing until the pump runs**; the public warning firing once and naming the host; and `quiet=0` as well as `quiet=1`.

Run locally under the embedded interpreter: `testing/tests/msa` + `testing/tests/predict` = 269 tests, all green.


javierbq:
## Verified against the live public server

The suite mocks the network by design, so I also ran the real thing once — human ubiquitin (1UBQ, 76 aa, already public in the PDB), through the actual `cmd.msa_search` path against `api.colabfold.com`:

```
 msa_search: api.colabfold.com is a PUBLIC, third-party MSA server. The query sequence
 LEAVES THIS MACHINE ...
 msa_search: 76 residues sent to https://api.colabfold.com [mode env]; search
 msa-3e19c90bf077 runs in the background and lands as ubq_live.
 msa_search: -> ubq_live, 20701 sequences x 76 columns from https://api.colabfold.com

source: {'kind': 'search', 'server': 'https://api.colabfold.com', 'mode': 'env',
         'ticket': 'teaD7s9Zgrc4P3Yj_z16v1V0UDxavt14ZtNuSw',
         'databases': ['uniref.a3m', 'bfd.mgnify30.metaeuk30.smag30.a3m'],
         'when': '2026-08-15T22:17:55', 'cached': False}
```

- ticket → poll → download → uniref+env merge works end to end; the stored query row is byte-identical to the input, and the merged a3m starts `>101` with no `#` cardinality line (which is what ColabFold's own merge assumes).
- provenance records the server, mode, ticket and both contributing databases.
- the public-server warning fired, under `quiet=0`.
- the second search of the same query was a cache hit with no submit (`cached: true`).

Read-only probes of the real server, run through the client code rather than curl:

| probe | live response | client behaviour |
|---|---|---|
| `GET /queue` | `{"queued":0}` | reachable |
| `GET /ticket/<well-formed unknown>` | 200 `{"status":"UNKNOWN"}` | → "does not know this ticket any more; it may have expired" |
| `GET /ticket/does-not-exist` | 400, **plain text** `invalid ID` | → "refused the request: HTTP 400 Bad Request" (non-JSON error body handled at the HTTPError branch, not misreported as "did not answer with JSON") |

**Not covered live:** the server already had a cached job for ubiquitin, so this completed in ~6 s — a genuinely multi-minute `RUNNING` poll, and the RATELIMIT/MAINTENANCE branches, are exercised only in the mocked tests. And nothing here has touched the private deployment: `colabfold.cluster.ziptx.internal` is VPC-internal, so a smoke test there is still worth doing before anyone relies on the private path.

Sidenote for #297: this alignment is 20 701 rows, above boltz-mlx's 16 384 admission limit, so truncation is a live concern for the consumer ticket rather than a theoretical one.

javierbq:
## Verified against the private deployment (`colabfold.accipiterbio.internal`)

The premise of the ticket — same binary, only the base URL changes — now checked rather than assumed. Driving the real `cmd.msa_search` path:

```
 msa_server: searches will use http://colabfold.accipiterbio.internal
resolve -> ('http://colabfold.accipiterbio.internal', 'msa_server')
is_public -> False                      # no third-party warning: nothing left the network
 msa_search: -> ubq_private, 20986 sequences x 76 columns
source: {'kind': 'search', 'server': 'http://colabfold.accipiterbio.internal', 'mode': 'env',
         'ticket': 'teaD7s9Zgrc4P3Yj_z16v1V0UDxavt14ZtNuSw',
         'databases': ['uniref.a3m', 'bfd.mgnify30.metaeuk30.smag30.a3m'], ...}
```

Plain `http` is accepted by the URL check, and the two error probes behave identically to the public server (`UNKNOWN` ticket → "does not know this ticket any more"; malformed id → HTTP 400 plain-text body → "refused the request").

### The cache key has to include the server, and here is the proof

Same query, same mode, **same ticket id** on both deployments — the backend hashes query+mode+databases — but different results, because the two carry different database vintages:

| | `api.colabfold.com` | `colabfold.accipiterbio.internal` |
|---|---|---|
| ticket | `teaD7s9Zgrc4P3Yj_z16v1V0UDxavt14ZtNuSw` | same |
| depth | 20 701 × 76 | 20 986 × 76 |

Keying the local cache on the query alone would have served one server's alignment for the other's.

### The RUNNING poll and a live cancel

Both earlier runs completed at submit (ubiquitin was pre-hashed on both servers), so these were re-run with novel scrambled-ubiquitin queries against the private deployment:

```
1. phase submit at 0.0s -> phase search at 2.0s -> phase ready at 34.3s
   done, 18971 sequences x 75 columns, query byte-identical to the input
   phases seen: ['submit', 'search', 'ready']

2. cancel while phase == 'queued' (ticket ueFKkBhEdmOXtG6VQa5buloez0XPDBrMcr8aTQ):
   state after cancel: cancelled          # immediately, not one poll interval later
   alignments unchanged: True             # and still nothing 5s + another pump later
```

So the poll loop, the phase reporting the app's marker carries, and the cancel-leaves-nothing-behind guarantee all hold against a real server.

**Still only covered by mocked tests:** RATELIMIT and MAINTENANCE. The private deployment has no rate limit configured and I am not going to hammer the public one to provoke a 429; the 429 handling is pinned by `testRateLimitSurfacesAsAnErrorNamingTheServer` against the exact body `backend/server.go` marshals.

PR #303, opened 2026-08-16
feat(msa): show a search in flight, where its alignment will appear

javierbq:
Answers "when the MSA is computing, is there any UI component indicating this?"
— there wasn't one. Stacked on #302, which the four new tests live in.

## The gap

An MSA search takes **minutes** and produces nothing until it lands. The
ALIGNMENTS section is fed by the alignment *store*, which a search does not
enter until it finishes — and the section hides itself when empty. So a user's
first `msa_search` showed **no sign anything was happening at all**.

The only thing reaching the app was the `MSA:` marker, and nothing consumes it:
`pollFeedback` has branches for `WEIGHTS:`, `PREDICT:`, `OBJPANEL:` and `MCP:`,
but not this one — so both lines a search emits fell through to the terminal
`else` and were appended to the console verbatim as raw JSON:

```
MSA:{"id":"msa-1351...","name":"prot_msa","state":"running","phase":"submit",...}
```

Its two siblings are properly surfaced; this was the odd one out:

| Operation | Marker | Handler | UI |
|---|---|---|---|
| Weight download (#286) | `WEIGHTS:` | `parseWeightsFeedback` | progress sheet |
| Prediction (#224) | `PREDICT:` | `BoltzJobManager` | placeholder + tooltip |
| MSA search (#298) — **before** | `MSA:` | none | none |
| MSA search — **after** | — | panel poll | row in ALIGNMENTS |

## Approach

Carried on the object panel's **existing 500 ms poll**, not by teaching Swift
the marker. The marker fires exactly twice — worker start and worker settle,
never in between — so a row driven by it could show neither the phase changing
nor the time passing. The poll already calls `msa.pump()`, so the tick that
lands an alignment also stops reporting the search: a row is never shown as
both a search and a result.

`searching.active()` already existed, documented as *"Drives the app's progress
row"*, with **zero callers**. This is its consumer.

## Design notes

- **No percentage.** The ColabFold server reports queue position and nothing else, so a bar would have to be invented — the same reasoning `Search.snapshot` already gives for reporting elapsed time instead.
- **In the ALIGNMENTS section, not a banner or sheet.** `msa_search` returns immediately and the user is expected to keep working, so a modal would be wrong; the row turning into the alignment it produces is the clearest statement of what happened.
- **Cancel by search ID, not name** — the name is not unique until the alignment lands.
- **The age is quantised to 5 s** (`SEARCH_ELAPSED_BUCKET`). The panel repaints when this payload changes and the poll runs at 2 Hz for minutes, so the raw float would repaint the section 120×/min to move a digit the row does not show.
- **`msa_searches` is optional Swift-side**, like every other field there: that decode is a single `guard let`, so one unknown key would not cost the alignments section — it would freeze the whole object panel on its last list.

## Tests

Four added to `msa_e2e.py` (from #302):

- a search in flight gets a row while `alignments` is still empty, and is replaced by its result in **one** tick;
- the row carries every field `MSASearchSummary` requires, read out of the Swift source — the same contract check #302 introduced, and it matters more here because those fields have no `CodingKeys`, so the JSON names *are* the property names;
- the age is quantised;
- no search means an empty list, not a missing key.

## Verification

- Python: **378 tests OK** — `msa` + `predict` + `raymol` + the two existing appkit object-panel suites, which validate the payload shape and are unaffected.
- Swift `UnitTests_macOS`: **256 tests, 0 failures** (4 env-gated skips), compiles clean.

Not verified in a running app: the installed RayMol.app predates the MSA PRs, so I could not drive the panel live. The row's *rendering* is unexercised by tests — the payload, the contract and the lifecycle are.

### DIFF
--- a/modules/pymol/appkit_inspector.py
+++ b/modules/pymol/appkit_inspector.py
@@ -476,6 +476,51 @@ def _alignment_map():
         return {}
 
 
+#: Seconds of granularity for a search row's age. The panel repaints when this payload
+#: changes, so sending the raw float would repaint it on every 500 ms tick for the
+#: several MINUTES a search runs. A bucket makes the row change 12 times a minute
+#: instead of 120, and it costs nothing that is visible: the row shows a coarse age
+#: because the server reports no progress fraction, and inventing one would be a
+#: plausible-looking lie (see Search.snapshot).
+SEARCH_ELAPSED_BUCKET = 5
+
+
+def _search_map():
+    """In-flight MSA searches (#298), oldest first, for the panel's progress rows.
+
+    An alignment takes MINUTES to build and does not exist until it lands, so without
+    this the object panel shows nothing at all while one is being searched for -- and
+    the ALIGNMENTS section hides itself when empty, so a user's first search gives no
+    sign anything is happening. `searching.active()` was written for this ("Drives the
+    app's progress row") and had no consumer until now.
+
+    Never raises, for the same reason _pending_map() and _alignment_map() do not: the
+    caller's single `except` writes no file at all, so a throw here freezes the panel
+    on a stale list.
+
+    Cheap by construction, and O(searches) rather than O(objects): `active()` is
+    normally EMPTY, and a snapshot is a handful of scalars copied under a short lock.
+    Nothing here reads the network or touches an a3m.
+    """
+    try:
+        from pymol.msas import searching
+        rows = []
+        for search in searching.active():
+            snapshot = search.snapshot()
+            elapsed = snapshot.get('elapsed') or 0
+            rows.append({
+                'id': snapshot['id'],
+                'name': snapshot['name'],
+                'phase': snapshot['phase'],
+                'server': snapshot['server'],
+                'elapsed': int(elapsed // SEARCH_ELAPSED_BUCKET
+                               * SEARCH_ELAPSED_BUCKET),
+            })
+        return rows
+    except Exception:
+        return []
+
+
 def poll_panel():
     """Write the object-list JSON to a temp file and print a short marker.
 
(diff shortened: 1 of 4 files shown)


javierbq:
Verified in a disposable macOS VM (Tahoe guest, Debug build of this branch), driving a real search against a fake ColabFold server stood up inside the app at the `pymol.msas.colabfold._urlopen` seam — so the worker thread, the phases, the main-thread pump and the panel poll all ran for real, with nothing leaving the VM.

**Search in flight** — the state that previously had no UI at all:

```
∨ ALIGNMENTS 1
     ⟳  barnase_demo_msa                    searching · 1m 30s
```

**The same row once it landed**, replaced in place:

```
∨ ALIGNMENTS 1
        barnase_demo_msa    barnase_demo/A         4 × 24
```

Note the section header counts the search (`ALIGNMENTS 1`) while it is still running — before this branch the section was hidden outright, which was the actual bug.

A second search run from the app command line (`msa_search barnase_demo, name=fresh_aln, refresh=1`) landed alongside it as `ALIGNMENTS 2`, unattached, confirming the attached and unattached rows render differently.

**One thing the snapshots caught that the tests could not.** With the row in place, the console still printed the `MSA:` marker verbatim — four wrapped lines of raw JSON per search, immediately below the row now saying the same thing in words. `pollFeedback` had no branch for it, so it fell through to the terminal `else`. Fixed in 5934a06: swallowed, like `MEASURE_ERR:`. Re-verified on both paths the marker fires on — a cache hit and a live `refresh=1` search — and the console now carries only:

```
 msa_search: 24 residues sent to https://msa.internal.example [mode env]; search
 msa-74b2a58be45f runs in the background and lands as fresh_aln. Stop it with "msa_cancel …".
 msa_search: -> fresh_aln, 4 sequences x 24 columns from https://msa.internal.example
```

Swift `UnitTests_macOS` still 256 tests / 0 failures after the change. VM released.

Issue #306, opened 2026-08-16
Add SimpleFold as a second predictor — the first that is not Boltz, so the request must start naming its runtime


javierbq:
## Summary

Add Apple's **SimpleFold** as a second prediction method, via the existing MLX-Swift port in
[`javierbq/simplefold-ios`](https://github.com/javierbq/simplefold-ios).

The Python side is close to the "one module and one line" that `docs/predictors.md` promises.
The Swift side is not: `BoltzJobManager` **is** the runtime, not a dispatcher to one, and the
request JSON `host.py` writes never says which predictor it came from. SimpleFold is the first
method that is not Boltz, so it is the one that forces that seam to exist.

## Settled scope

Two constraints are decided, not open questions, and the rest of this issue is written against
them:

- **macOS only.** Same platform filter BoltzMLX already has in `project.yml`. See "Platform" below.
- **Single chain only.** SimpleFold cannot model complexes. More than one chain is a **plain
  error** — `parse_spec` raises and nothing is submitted. Not a warning, not a best-effort fold.
  See "Rejecting complexes" below.

## Why SimpleFold

It is a genuinely different method, not another weight pack — which is what makes it worth the
seam, and what makes `boltz2-bf16` the wrong precedent to reason from (that one overrides
nothing but `weight_bundle`).

- **Single-sequence by construction.** ESM2 embeddings + a flow-matching FoldingDiT sampler; no
  MSA anywhere in the pipeline. So `supports_msa = False` is not a stub — it is the accurate
  declaration, and the base class then refuses `predict ..., msa=...` **by name** rather than
  folding single-sequence and saying nothing. #274's design gets its first real exercise here:
  today every registered predictor supports MSAs, so the refusal path is only ever hit by tests.
- **No trunk, no recycling, no reverse diffusion.** Its one quality knob is the number of
  flow-matching timesteps (500 in the port). Every knob `cmd.predict` currently names is
  meaningless to it — see "Options" below.
- Fast single-sequence folding is the case Boltz-2 serves worst, so this is additive to users,
  not a variant.

## What already exists

`javierbq/simplefold-ios` — MLX-Swift port derived from
[`apple/ml-simplefold`](https://github.com/apple/ml-simplefold), MIT.

- A SwiftPM package: `.library("SimpleFoldBenchCore")`, `platforms: [.macOS(.v14), .iOS(.v17)]`.
[...]

PR #307, opened 2026-08-16
feat(predict): the runtime seam a second predictor needs, and the knob bug that blocked one

javierbq:
#306 is **closed** — Apple's `LICENSE_MODEL` restricts the SimpleFold weights to non-commercial research and excludes product use, so RayMol cannot serve the pack as a `WeightBundle` the way `boltz-mlx` does. See [the closing comment](https://github.com/javierbq/RayMol/issues/306#issuecomment-5309744565) for the full finding.

This PR is the part of that work which was never SimpleFold-specific, and stands on its own.

## A latent bug: a non-Boltz predictor was impossible to call

`predicting.py` put `recycling_steps` and `diffusion_steps` into `requested` unconditionally, so `validate_options` rejected them **by name** for any predictor that had not declared them — including on a call that named no knobs at all. Any method without a trunk was uncallable. Both are conditional now, as `msa_depth` already was; the defaults live in each predictor's `option_defaults`, which is the thing that knows them.

## The runtime seam

`BoltzJobManager` *is* the Boltz backend, not a dispatcher to one, and the request named no runtime.

- `host.submit` writes `runtime`; `preflight` refuses one this build does not carry. Weights and featurizer are method-specific, so running another method's request on the Boltz backend would not fail — it would return a confident wrong structure.
- Optional at the far end, absent meaning `boltz`, so an older Python side still decodes.
- `PyMOLBridge` advertises what is linked in `RAYMOL_PREDICT_RUNTIMES`; `host.require_runtime` reads it, so `check_available` can refuse *before* a weight download instead of after.
- Each predictor puts only its declared knobs on the wire (`knobs=self.option_defaults`). A Boltz request is unchanged.

## The simplefold module

Kept, registered, and inert: it validates input and refuses with "no published weight pack yet." It is the concrete exercise that proved the seam, and it documents the method's real constraints in code — single chain only (the port would fold two chains as their concatenation and emit a plausible-looking PDB), canonical 20 only (its tokenizer maps unknown letters to X), and `supports_msa = False`.

Say so if you would rather not carry a predictor that may never run, and I will strip that module and its tests, leaving the seam and the bug fix.

## Behaviour change worth a look

`predict <Tab>` now lists the three ids instead of silently inserting `boltz2`. While every id began with `boltz2` the completer had a common prefix to fill in; an id from another family removes it. `predict s` completes to `simplefold, ` in one keystroke. Two completion tests encoded the old two-Boltz world and were updated.

## Verification

- 36 new Python tests, 4 new Swift tests. `testing/tests/predict` 263 green, `testing/tests/msa` 127 green — the latter includes the source-reading test that pins Python's wire keys against Swift's `CodingKeys`, which caught the missing `runtime`/`num_steps` decode.
- Full Swift unit suite green; macOS slice compiles.
- **Exercised in the real app**: a Boltz-2 fold of villin HP36 ran end to end through the changed wire (request carried `runtime: "boltz"` and no `num_steps`), finished in 25.5 s at 620 MB peak, and auto-loaded with the seed in the state title and real pLDDT in the B-factor — mean 96.2, correct three-helix bundle. The refusal paths were checked live too.
- **iOS slice not verified**: it fails identically on master with an AppIcon asset-catalog error, before any Swift compiles. This change is invisible to that slice — both edits sit behind `#if os(macOS)` / `#if TARGET_OS_OSX`.

### DIFF
--- a/docs/predictors.md
+++ b/docs/predictors.md
@@ -10,7 +10,7 @@ it is shaped this way.
 
 | File | Responsibility |
 |---|---|
-| `modules/pymol/predicting.py` | the `cmd.*` surface; you should not need to touch it |
+| `modules/pymol/predicting.py` | the `cmd.*` surface; touch it only to NAME a new knob (see step 6) |
 | `modules/pymol/predictors/base.py` | `Predictor` contract, `PredictionSpec`, `PredictionOptions` |
 | `modules/pymol/predictors/weights.py` | `WeightBundle`, `BundledSource`, `WeightCache` |
 | `modules/pymol/predictors/registry.py` | `register` / `get` / `available` / `unregister` |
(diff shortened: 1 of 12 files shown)


javierbq:
closing due to license issues

PR #310, opened 2026-08-17
Object metrics: numbers a tool measured, kept with the object it measured (#308)

### LINKED ISSUE #308: Objects can carry no data: RayMol needs generic per-object metrics, scoped to the object, a state, or a residue range
# Objects can carry no data: RayMol needs generic per-object metrics, scoped to the object, a state, or a residue range

## Summary

Every tool RayMol has grown produces *numbers about an object* — and the object has nowhere to
put them. A prediction reports per-residue confidence and pairwise error; a design pass reports
native fit, certainty and a per-residue propensity; an MSA search reports depth and coverage;
`align` reports an RMSD and an atom count; `assign_stereo` reports an assignment per centre.
Today each of these is printed, smeared into the B-factor column, cached in Swift, or written to
a temp file keyed by a job id — and all of it is gone when the session closes.

This asks for **generic metric machinery**: a tool declares what it measures, at what scope, and
writes it against an object; the store, the panel, `spectrum`, and export handle it without
knowing which tool it came from. No per-tool special cases, no new code in the store when the
fifth predictor lands.

The design problem worth arguing about is not storage — it is **scope**. A number is not simply
"about an object": mean pLDDT is about one *model*, sequence recovery is about the *sequence* and
identical across all five models, an RMSD is about a *pair* of objects, and a confidence array is
about *residues*. Getting that wrong is how a metric silently ends up describing something it
does not describe.

## Scope is the design

A metric declares its scope; the store never infers one.

| Scope | Meaning | Examples |
|---|---|---|
| `object` | True of the object as a whole — typically sequence-derived, identical across states | sequence recovery, MSA depth used, chain count, tool provenance, wall-clock |
| `state` | True of one model / one set of coordinates | mean pLDDT, clash score, radius of gyration, RMSD to a reference, energy |
| `chain` | Per chain, at either of the above scopes | per-chain pLDDT, per-chain identity, ipSAE for a chain pair |
| `residue` | An array indexed by `(chain, resi)` | per-residue confidence, design certainty, SASA, per-residue RMSD |
| `pair` | An array indexed by `(chain, resi) × (chain, resi)` | predicted aligned error, contact probability, coupling |

Two rules make the scope carry its weight:

- **A scope mismatch is an error, not a coercion.** Writing a `state`-scope metric with no state,
  or a `residue` array whose indices are not in the object, is refused at write time — the same
  discipline `Predictor.validate_options` already applies to inference knobs, and for the same
  reason: the alternative is a number that reads fine and means something else.
- **The store never aggregates.** A tool may declare that its `residue` array summarises to a
  `state` scalar by mean, and it may write both — but the store computes neither. Re-deriving
  "mean pLDDT" from a rounded B-factor column is exactly the lossy path this issue exists to
  close.

`n_models=5` is the case that forces this. Five full runs land as five states of one object.
Their pLDDT arrays are `state`+`residue`. The sequence they folded, the MSA and its depth, and
the recovery against it are `object` — one value, not five copies.

## The record

Metrics are grouped by the **run** that produced them, because a metric without its inputs is not
evidence of anything:

- provenance: tool id, tool/weight version, timestamp, the validated option dict verbatim, the
  seed, the input selection or sequence, the alignments and depth actually used
- the metrics themselves: `key → (scope, value | array, units, index convention)`
- what it produced: which states of which object

An object carries **many runs** — fold, re-fold with a deeper MSA, then design — and no run
replaces another. Never an `object → metrics` dict.

## A tool declares its metrics once

The point of the machinery. A tool registers a small schema — per key: scope, dtype, units,
expected range, whether higher is better, a display label, and an optional summarisation rule.
Everything generic then falls out of that declaration, for tools that do not exist yet:

- the object panel renders a run's summary row without knowing the tool
- `spectrum` can colour by any `residue`-scope metric by name, instead of by whatever happens to
  be sitting in `b`
- export to JSON/CSV is one code path
- objects can be sorted or filtered by an `object`-scope metric
- two runs of the same tool are comparable because they share a schema

Commands mirror the `msa_*` family and register the same way in `keywords.py`: `metrics_list`,
`metrics_get`, `metrics_delete`, `metrics_export`, `metrics_load` (for a metric file computed
outside RayMol), and a `metrics_color` that drives `spectrum` from a stored array. Naming is the
one thing I would like a second opinion on — `metrics_*` says what these are, `data_*` is shorter
and vaguer.

Storage follows `pymol/msas/store.py`: process-wide dict, `.pse` round trip under its own
version-gated session key, arrays gzip+base64. Its header argument applies unchanged — this has
no geometry, so making it a `CObject` would stub out every virtual and put a row in the object
list that accepts `show cartoon` and lies.

## What this would capture that is thrown away today

| Tool | Produces | Where it goes now |
|---|---|---|
| boltz2 / boltz2-bf16 | per-residue pLDDT; full PAE matrix; min ipSAE / ipAE | pLDDT → B-factor column; **PAE and `interfaceScores()` are computed and dropped** — `BoltzJobManager.swift:475` reads only `plddt`, and `InterfaceScores` appears nowhere in `swiftui/PyMOLViewer/` |
| simplefold | a different metric set — no trunk, no diffusion | nothing |
| ProteinMPNN | `nativeFit`, `certainty`, 20-way propensity per residue | Swift-side `DesignScoreCache`; per-residue score into `p.mpnn_conf`, falling back to `b` on open-source builds (`raymol_design.py:227`) |
| msa_search | depth, columns, coverage, server | on the MSA object — but **which run used it is recorded nowhere** |
| any predictor | `elapsed_s`, `peak_bytes`, options, seed | temp status file keyed by job id (`host.py:107`); only the seed reaches the object, as a state-title string (`predicting.py:345`) |
| `align` / `super` / `cealign` | RMSD, atoms aligned, cycles | returned and printed |
| `assign_stereo` | per-centre assignment | `p.stereo`, incentive-only, with the same fallback problem |

Two of these collide *right now*: pLDDT lands in `b`, and a design pass on that object overwrites
`b` with its own score, with nothing in the session saying the column changed meaning.

## Non-negotiables

- **Scope is declared and enforced.** See above; this is the whole point.
- **`b` becomes a view, not the storage.** `spectrum b` keeps working, but the record is
  authoritative and the column is derived from it on request — so two tools can coexist on one
  object, and either metric can be re-applied later.
- **Absent is not zero.** A single-chain fold has no interface score; a masked residue has no
  design score. Design already models this correctly — `value: null` rows are excluded rather
  than clamped to the `-9999` sentinel (`raymol_design.py:186`) — and that must survive the round
  trip.
- **Arrays are stored as the tool produced them**, with an explicit `(chain, resi)` index rather
  than a bare offset. A structure with unobserved residues is precisely where a positional array
  goes wrong, and it is the same mismatch `_check_alignment_query` already exists to catch.
- **An unreadable or future-versioned payload is skipped, not guessed**, and a `.pse` carrying
  metrics still opens in upstream PyMOL. The #296 rule.
- **The panel must not compute.** The object panel polls the main thread every 500 ms; a row shows
  precomputed summary scalars and never touches an array. Same discipline as `MSA.summary()` and
  the fix in #271.
- **No per-tool branches in the store.** If adding a predictor needs an edit inside the store,
  the schema declaration is not carrying enough.

## Open questions

- **Binding to states.** A state is a positional index — delete or reorder states and a
  `state`-scope metric now describes different coordinates. Options: bind by index and invalidate
  on any state-count change, or stamp a per-state identity into the title (the seed already
  travels this way). I lean towards the former, loudly.
- **Rename, delete, `create`.** The MSA store resolves its target lazily by name, so a rename
  orphans it. Metrics should follow `set_name` (`editing.py:447`); dropping on delete is obvious;
  but a `create` from a subset selection should almost certainly *not* inherit metrics whose
  residue indices no longer apply.
- **Pair metrics.** An RMSD belongs to two objects. Recording it on the mobile object with the
  reference named in provenance is the cheap answer, and is probably right — a `pair` scope for
  whole objects earns its complexity only if something else needs it.

## Suggested split

1. **The core.** Store, scope model, schema registry, `.pse` round trip, `metrics_*` commands.
   Tool-agnostic, with no predictor code touched.
2. **Predictions as the first adapter.** Emit metrics beside the result PDB from the Swift host
   (PAE, pLDDT, `interfaceScores()`, elapsed, peak), extend the status/result contract in
   `host.py`, and bind per-state records in `deliver_result` — including per-model under
   `n_models`.
3. **Design as the second adapter**, which is the real test of genericity: a different scope mix,
   a different metric set, and it stops owning `b`.
4. **Surfacing.** Panel section, `metrics_color`, export. A PAE/pLDDT plot view is its own issue.

## Out of scope

- Plot and matrix UI beyond a summary row.
- Per-atom custom properties. `p.*` is incentive-only and open-source raises
  `IncentiveOnlyException` — which is why both Design and `assign_stereo` already carry fallbacks.
  This store sits beside that mechanism rather than depending on it.
- Cross-session metric history or an on-disk database. A record belongs to a `.pse`.

javierbq:
Closes #308. Branched from `master`; contains only the metrics work.

Every tool RayMol has grown produces numbers about an object, and the object had nowhere
to put them. Each was either never read out of the runtime (boltz-mlx computes a full PAE
matrix and `interfaceScores()` on every scored run — both were dropped on the floor),
smeared into the B-factor column, or left in a temp status file keyed by a job id that
dies with the process.

This adds the missing noun: a named, session-stored **run** of measurements attached to an
object, generic across tools.

## Scope is the design

A metric declares one of `object` / `state` / `chain` / `residue` / `pair`, and it is
checked when written. `n_models=5` is what forces the distinction: five models each have
their own `mean_plddt`, while the sequence they folded is one value for the object and the
alignment depth is one value per chain. A number at the wrong scope reads perfectly and
describes something else, so a scope mismatch is an error, not a coercion — an
`object`-scope value carrying a state, a state that does not exist, an array indexed by
residues the object does not have are all refused, naming the offender.

The store never aggregates. A spec may *declare* that an array summarises by mean and a
tool may write both; which residues went into a mean is a property of the tool.

## Generic by construction

A tool declares what it measures once, and the store, the object panel, `metrics_color` and
export work off that declaration — the next predictor needs no edit in the store.
Predictors declare theirs as `Predictor.metric_specs`, and only what they genuinely
produce: `predictors/metrics.py` carries the shared sets (`SCORED_SPECS` for a method with
a confidence head, `UNSCORED_SPECS` for one without), the same capability contract
`supports_msa` already carries.

## `b` and `p.*` are views, not records

Recording happens before colouring, and the B-factor column is treated as a rendering
channel the store can refill. Two corrections to how I first justified this — CI caught
the second:

- **RayMol implements `p.*` atom properties** (`layer1/Property.cpp`, ungated), so a
  design pass writes `p.mpnn_conf` and does *not* clobber a prediction's pLDDT in `b`.
  That collision is real only on a build without properties — stock open-source PyMOL —
  where `apply_design_coloring` falls back to `b`.
- **Neither channel is a record either way.** A property is per atom: no run, no units,
  no provenance, no per-state or per-pair value, and it does not survive a `.pse`. `b`
  is one unlabelled scalar per atom, rounded by any PDB round trip, displaced by the
  next thing that colours through it — `metrics_color` included.

So `metrics_color` can put any stored array back at any time, and the test asserts the
build-independent part: both runs are in the store, and the confidence re-applies.

## What's here

| | |
|---|---|
| `modules/pymol/metrics/` | store, schema registry, session-aware scope checks, document format, `.pse` round trip |
| `modules/pymol/metric.py` | `metrics_list` / `get` / `color` / `export` / `load` / `delete` / `schema` |
| `BoltzJobManager` | writes pLDDT, PAE, mean pLDDT and the interface scores beside the result; the residue index mirrors `StructureWriter`'s per-chain numbering, or arrays land a register out |
| `ObjectPanel` | a METRICS section — scalars only, since the 500 ms poll must never walk a PAE matrix |
| `docs/metrics.md` | user guide, with a tool-author half |

Absent stays absent throughout: a single-chain fold has no interface score and a masked
residue has no design score, and `0.0` there reads as a terrible measurement rather than as
none.

Session behaviour follows the alignment store (#296): runs ride the `.pse` under
`raymol_metrics` with arrays gzipped, a session carrying none empties the store,
`reinitialize` empties it, a run from a tool this build lacks is still restored, `set_name`
moves runs and `delete` drops them, `create` does not inherit them. States are positional,
so a changed state count flags the run stale rather than silently re-pointing it.

## Ambiguity is refused, not guessed

Recency only orders runs that describe the same thing. Two cases where it does not, both
refused with the disambiguator named:

- several **tools** measured the object — `tool=` picks one;
- several **states** were measured, which is what `n_models` produces — `state=` picks one.

`run=` always wins outright, and a key only one tool declares needs nothing.

## Verification

- 102 new Python tests; `predict` (226) and `msa` (127) unchanged.
- macOS app builds clean; 256 Swift unit tests pass (`UnitTests_macOS`).
- Every command exercised through `cmd.do` at `quiet=0`, which is where two display bugs
  and the `state=` resolution bug were found.
- **One real boltz2 fold on device**, end to end — see the comment below: the host-written
  document parsed into a run with pLDDT, a 35×35 PAE matrix and `mean_plddt=88.66`, with the
  stored array matching the PDB's B-factors to two decimals.
- `testing/tests/metrics` added to the embedded-tests workflow.

## Follow-ups, deliberately not here

- **`simplefold` declares `metric_specs = UNSCORED_SPECS`** — eight lines that belong with
  #307, since that is where the predictor is added. Only coupling this branch had to it.
- **`protenix-base`** builds on both #307's runtime seam and this branch's
  `CONFIDENT_SPECS`, so it stacks on top of both rather than riding along here.


javierbq:
## End-to-end verified on device

The one gap called out in the description — the metrics document had never been round-tripped
through a real prediction — is now closed. Not in a VM: MLX has no Metal there, so the fold
could never run. Done on the host, with a Debug build of this branch driven by a temporary
startup script (its MCP server disabled so it left the running instance's handoff alone).

**One real boltz2 fold**, 35-residue helical peptide, seed 7, warm weight cache:

```
status     state=done  elapsed_s=24.33  peak_bytes=594482496
run        boltz2_fe3e9cab  tool=boltz2  object=pep  version="boltz2-mlx-int8 v1"  states=[1]
keys       n_residues, n_chains, elapsed_s, peak_bytes, plddt, mean_plddt, pae
scalars    n_residues=35  n_chains=1  mean_plddt=88.66
inputs     seed=7, options={recycling_steps:3, diffusion_steps:200, msa_depth:16384, ...}
```

The host document `BoltzJobManager.writeMetrics` actually wrote (25 KB):

```
plddt       state=0  index=35  values=35
mean_plddt  state=0  value=88.65866273478824
pae         state=0  index=35  values=1225        # 35² — computed and dropped before this PR
```

What the run confirms:

- **PAE survives.** 1225 values over a 35-residue index, `min 0.25 / max 29.96` Å.
- **The index lines up with the structure.** `A/1 … A/35`, and the B-factors PyMOL parsed
  out of the PDB match the stored array to the PDB's own two decimals
  (`74.24, 86.06, 88.74, 90.26, 93.03` vs `74.2368…, 86.0551…, 88.7437…`). That is the
  `residueIndex`-mirrors-`StructureWriter` claim checked against real output rather than by
  reading both functions.
- **State is stamped on this side.** The document says `state=0`; the run carries `state=1`,
  the state the model landed in.
- **No double-sourcing.** `elapsed_s` and `peak_bytes` came from the status file, not the
  document.
- **Absent stays absent.** No `min_ipsae`/`ipae` (single chain — undefined, not 0) and no
  `msa_depth` (single-sequence fold).

Then the same bytes through the rest of the path: `metrics_load` → `metrics_color plddt`
(35 residues, B-factors as above) → `.pse` save → `reinitialize` (store empties) → reload →
1225 PAE values over 35 residues and `mean_plddt=88.66` intact, in a 188 KB session.

Host environment restored afterwards: Debug instance quit, temporary `~/.raymolrc.py`
removed, MCP handoff file restored, VM lease released.

Issue #309, opened 2026-08-17
Add Protenix as a predictor via protenix-mlx — the first method whose featurizer does not exist in Swift


javierbq:
## Summary

Add **Protenix** (ByteDance's AlphaFold3-class predictor) as a prediction method, via the
MLX-Swift port in [`javierbq/protenix-mlx`](https://github.com/javierbq/protenix-mlx).

The port is further along than simplefold-ios was at #306: the network is complete and
verified against PyTorch to 2e-4, it folds a sequence end to end in Swift, and **twelve
weight packs are already published as digest-pinned release ZIPs whose members are exactly
`config.json` / `manifest.json` / `model.safetensors`** — the same three-member shape
`boltz2`'s bundle declares, so `WeightCache` needs no change at all.

But it is blocked on something neither Boltz nor SimpleFold has: **its featurizer is not
in Swift.** `ProtenixPredictor.fold` takes a `FeatureBundle` that `protenix-mlx
export-features` produces in Python — driving upstream Protenix, with torch and the CCD
cache present. Nothing in RayMol can run that. Deciding what to do about that is most of
this issue.

## What already exists

`javierbq/protenix-mlx` — Apache 2.0, derived from
[`bytedance/Protenix`](https://github.com/bytedance/Protenix), same shape as boltz-mlx: a
Python exporter, a `.library("ProtenixMLX")` SwiftPM product, packs distributed as release
ZIPs pinned by digest.

| Layer | State |
| :--- | :--- |
| Python exporter (int8 / fp16 / bf16), artifact schema, packing | working, verified against real checkpoints |
| Swift core layers (Transition, AdaLN, AttentionPairBias, TriangleMultiplication ×2, TriangleAttention ×2, OuterProductMean, MSAPairWeightedAveraging) | working, verified vs PyTorch |
| Swift trunk (input embedder, relpos, MSA module, Pairformer, recycling) | working, verified |
| Swift diffusion (atom encoder/decoder, transformer, conditioning, sampler) | working, verified |
| Swift end-to-end fold (feature bundle → coordinates → PDB) | working |
| **Featurization in Swift** | **does not exist** |
| Confidence / distogram heads | not implemented (`Sources/ProtenixMLX/Confidence/` is empty) |
| Templates, real MSA | not implemented |

37 Swift tests, 49 Python tests. Parity is by recorded fixtures that reproduce
bit-identically from their seed and **skip loudly** when absent, and it is mutation-verified:
[...]

PR #312, opened 2026-08-17
Add Protenix, and the runtime seam a second backend needs (#309)

javierbq:
Adds **Protenix** — ByteDance's AlphaFold3-class predictor — via
[`protenix-mlx` v0.1.0](https://github.com/javierbq/protenix-mlx/releases/tag/v0.1.0),
and with it the seam a request needs to name which backend runs it. Closes the Python
half of #309.

Python side only. No build carries the `protenix` runtime yet, so `check_available`
refuses before anything is downloaded.

## Why this PR carries the runtime seam

#307 drafted that seam for SimpleFold, which is not going to be integrated. Protenix is
now the first method that is not Boltz, so the seam lands here — the same design, minus
everything SimpleFold-specific:

- **`Request.runtime`**, optional on the Swift side, absent meaning `boltz`. A required
  field would turn any Python/Swift skew into "malformed prediction request" instead of a
  refusal naming the runtime — the reasoning `object_name` and `alignments` already follow.
- **`preflight` refuses an unknown runtime first**, before the size model, which is fitted
  to Boltz's measured peaks and would be meaningless applied to another method's request
  even as a refusal.
- **`RAYMOL_PREDICT_RUNTIMES`**, advertising what is actually linked, so a predictor whose
  backend is missing refuses in `check_available` rather than after a weight download.
- **`host.submit(knobs=)`**, so a request carries exactly the knobs its method declared.
  Protenix has no alignment, and `msa_depth` on the wire would record a run as having used
  one.

Not carried over: SimpleFold itself, its `num_steps` knob, and making Boltz's two knobs
optional — Protenix declares both, so no request omits them.

The reason any of this exists rather than dispatching on the predictor id: weights and
featurizer are method-specific, so a Protenix request reaching the Boltz backend would not
fail. It would tokenize with the wrong featurizer and return a confident wrong structure.

## The predictor

Accepts **complexes**, which the method #307 was written for could not: the port's
featurizer groups chains into entities and builds the cross-chain pair features, so a
multimer is modelled rather than approximated. Identical sequences become copies of one
entity, which is what tells the model a homodimer's chains are related. The residue cap is
on the total, since peak memory follows total tokens rather than the longest chain.

It is also the first predictor that **adds no knob to `predicting.py`** — it recycles a
trunk and runs reverse diffusion, so `recycling_steps` and `diffusion_steps` already mean
what it needs.

## MAX_RESIDUES is 400, and sits below the largest measurement

| residues | 60 | 120 | 250 | 400 | 550 | 700 |
|---|---|---|---|---|---|---|
| peak (MiB) | 547 | 942 | 2279 | 3868 | 6303 | 8622 |
| mean pLDDT | 82.7 | 47.3 | 30.9 | 27.2 | 26.2 | 26.3 |

700 residues runs, at 8.6 GB and six minutes. Alongside a loaded session that is where a
jetsam kill takes unsaved work, and the model's own confidence there is 26 — six minutes
to produce a structure it says nothing can be concluded from. 400 (3.9 GB) is the largest
length that both fits comfortably and is worth the wait.

Measured against **MLX's own high-water mark, not process RSS**. That distinction is not
pedantry: by RSS, 400 residues appears to cost *less* than 60, because MLX allocates
through Metal and recycles buffers in a cache it need not return to the OS. The first
sweep I ran was thrown away for this reason.

## What it deliberately does not declare

`metric_specs` is empty, even though the runtime computes pLDDT, PAE, PDE and resolved
(verified against PyTorch to 2e-4). A per-residue array needs the metric store from #308
to land in, and declaring keys with nowhere to write them would put numbers in the schema
that never arrive. The declaration follows #310.

`supports_msa` is `False` and staying so until an a3m can reach `msa_features`: the port
feeds the network upstream's depth-1 dummy alignment, and a method that accepted a real
alignment then folded against that dummy would return a worse structure with nothing
saying so.

## Single-sequence confidence collapses with length

Worth knowing before this ships to anyone. Mean pLDDT over **complete domains**, not
truncations:

| villin 35aa | GB1 56aa | ubiquitin 76aa | barnase 110aa | lysozyme 129aa |
|---|---|---|---|---|
| 94.8 | 72.3 | 67.5 | 58.9 | 37.0 |

This is the documented behaviour of an AF3-class model with no alignment, not a fault in
the port — the same pack scores villin 94.8 against poly-alanine 78.4 at equal length, and
all four confidence outputs match PyTorch to 2e-4. It means Protenix earns its two minutes
on small domains and not much else until #274 gives it a real alignment. Lysozyme is the
low outlier for a second reason: four disulfides, and `token_bonds` is all zeros here.

## Two adjustments this forces elsewhere

- **`predict_weights download=1`** with no id iterates every predictor without consulting
  `check_available`, so a 214 MB pack for an unlinked runtime would have been pulled by a
  command aimed at the predictors that work. A bulk fetch now takes only what the build can
  run; naming a predictor explicitly still fetches it.
- **`predict <Tab>` no longer completes.** `boltz2*` and `protenix-base` share no prefix,
  so the parser has nothing unambiguous to extend and returns `None`. Two tests in
  `predict_completion.py` had encoded the old registry's contents; they now assert the
  mechanism, plus a new one showing `predict p<Tab>` completes in one step — the payoff for
  not naming it a bare `protenix`, which would have created the same dead end `boltz2`
  already does for `boltz2-bf16`.

## Testing

285 predict tests pass, 58 of them new, including `predict_runtime.py` which tests the
seam directly rather than only through one predictor — the back-compat properties (absent
declaration means boltz, on both sides; every registered predictor names a runtime) are
exactly the kind that rot silently. 127 MSA tests still pass.

macOS app slice compiles. **The iOS slice fails** on a pre-existing missing `AppIcon` named
"RayMol" in `Assets.xcassets` — reproduced identically on `master` with these changes
stashed. Both files touched here are inside macOS-only guards (`#if os(macOS)`,
`#if TARGET_OS_OSX`), so nothing added reaches the iOS slice.

## Follow-ups

- The Swift `protenix` runtime — a `ProtenixJobManager`, per-runtime `PredictSizeGuard`,
  and the metrics document. Tracked in #309.
- `metric_specs` once #310 lands.
- Real MSAs (#274), which is what would make this method useful above ~130 residues.

### DIFF
--- a/docs/predictors.md
+++ b/docs/predictors.md
@@ -19,16 +19,39 @@ it is shaped this way.
 
 ## Shipped predictors
 
-| id | weights | notes |
-|---|---|---|
-| `boltz2` | affine-int8, 507 MB | the default |
-| `boltz2-bf16` | dense bfloat16, 996 MB | same model, same runtime, unquantized |
+| id | runtime | weights | notes |
+|---|---|---|---|
+| `boltz2` | `boltz` | affine-int8, 507 MB | the default |
+| `boltz2-bf16` | `boltz` | dense bfloat16, 996 MB | same model, same runtime, unquantized |
+| `protenix-base` | `protenix` | affine-int8, 214 MB | complexes, single sequence. **Not runnable yet** — no build carries the `protenix` runtime, see #309 |
 
-`boltz2-bf16` subclasses `Boltz2Predictor` and overrides nothing but `weight_bundle`.
-That works because `host.submit` sends `weights_dir` per job and the Swift runtime
-picks its matmul path from the artifact manifest — a dense pack declares no
-quantization block — so one runtime serves both with no host change. It needs
-boltz-mlx >= 0.2.0.
+## Runtimes
+
+A predictor's **runtime** is the Swift backend that runs it, named on the wire by
+`host.submit(..., runtime=)`. It is a separate axis from the predictor id: `boltz2` and
+`boltz2-bf16` are two predictors and one runtime, because `host.submit` sends `weights_dir`
+per job and the Swift side picks its matmul path from the artifact manifest — a dense pack
+declares no quantization block — so one runtime serves both.
+
+The distinction has to be on the wire because weights and featurizer are method-specific.
+Running one method's request on another's backend does not fail; it tokenizes with the
+wrong featurizer and returns a confident wrong structure. So:
+
+- `BoltzJobManager` implements `boltz` and **refuses** any other in `preflight`, before it
+  applies its size model — which is fitted to Boltz's measured peaks and would be
+  meaningless applied to another method's request even as a refusal.
+- `PyMOLBridge` advertises what is actually linked in `RAYMOL_PREDICT_RUNTIMES`, which is
+  what lets `check_available` refuse **before** a weight download rather than after.
+- `Request.runtime` is **optional** on the Swift side, absent meaning `boltz`. A
+  non-optional field would turn any Python/Swift version skew into "malformed prediction
+  request" instead of a refusal that names the missing runtime. The same reasoning applies
+  to `object_name` and `alignments`.
+
+That default is also the trap: a new predictor that forgets to pass `runtime=` is silently
+dispatched to Boltz. `predict_runtime.py` asserts every registered predictor names one.
+
+`boltz2-bf16` subclasses `Boltz2Predictor` and overrides nothing but `weight_bundle`, which
+is what "two predictors, one runtime" looks like in practice. It needs boltz-mlx >= 0.2.0.
 
 It is **not** an upgrade. On an M3 Pro at 117 tokens, recycling 3 / 200 steps, dense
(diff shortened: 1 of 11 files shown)


javierbq:
Superseded by #313, which now targets `master` directly and contains every commit here plus the Swift runtime, the six-pack table and the measured size guard.

Nothing is lost by closing this one — I have left it open rather than closing it myself, since it is the review record for the seam. Close it whenever you like, or if you would rather review in two passes, say so and I will split #313 back apart.

javierbq:
Superseded by #313, now merged as 66515c21f — every commit here landed in master through it, rebased. Closing as redundant rather than stale.

The seam this PR introduced (`Request.runtime`, `RAYMOL_PREDICT_RUNTIMES`, `host.submit(knobs=)`) is on master and now has a second runtime using it.

PR #313, opened 2026-08-17
Add Protenix: the second inference runtime, and six weight packs (#309)

### LINKED ISSUE #309: Add Protenix as a predictor via protenix-mlx — the first method whose featurizer does not exist in Swift

javierbq:
Adds **Protenix** — ByteDance's AlphaFold3-class predictor — to `cmd.predict`, via
[`protenix-mlx`](https://github.com/javierbq/protenix-mlx), and with it the seam a request
needs to name which backend runs it. Closes #309.

`predict protenix-base-int8, <sequence>` folds on-device and loads the scored structure.

> Supersedes #312, which is fully contained here — this PR now targets `master` directly.
> #307's runtime seam is included rather than depended on, since that PR is not landing.

## Verified in the app, not only in tests

- The 225 MB pack downloads through RayMol's own `WeightCache` in 17 s, digest and member
  set checked on the way in.
- A 35-residue sequence folds in **34.7 s at 490 MiB peak** and loads with per-residue
  pLDDT in the B-factor column.
- At a fixed seed the coordinates are **bit-identical to `ProtenixMLXCLI`** from the same
  pack — max deviation 0.0000 Å across 289 atoms. That is what says the integration adds
  nothing and corrupts nothing.

## The runtime seam

Protenix is the first method that is not Boltz, so the request has to say so. Weights and
featurizer are method-specific: a Protenix request reaching the Boltz backend would not
fail, it would tokenize with the wrong featurizer and return a confident wrong structure.

- **`Request.runtime`**, optional on the Swift side, absent meaning `boltz` — a required
  field would turn Python/Swift skew into "malformed prediction request" instead of a
  refusal that names the runtime.
- **`preflight` refuses an unknown runtime first**, before the size model, which is fitted
  to Boltz's peaks and would be meaningless applied elsewhere.
- **`RAYMOL_PREDICT_RUNTIMES`** advertises what is linked, so `check_available` refuses
  before a download rather than after.
- **`host.submit(knobs=)`** puts only the knobs a method declared on the wire; Protenix has
  no alignment, and `msa_depth` there would record a run as having used one.

The job shell is `BoltzJobManager`'s, reused rather than reimplemented. Routing lives in
`handle(marker:)` only because it owns the marker; a third runtime should lift that into a
dispatcher, and the code says so.

## Three upstream releases this needed

**0.1.1 — progress and cancellation.** `fold()` was synchronous with no hook, so
`predict_cancel` would have done nothing for up to six minutes. Now a handler fires per
trunk recycle and per diffusion step; returning `false` throws. Finer than the Boltz path,
whose trunk has no cancellation points at all.

**0.1.2 — a real bug, found by measuring output.** The sampler used struct defaults
(`step_scale_eta` 1.0, `gamma0` 0.0) instead of the pack's own 1.5 / 0.8 — values every
`config.json` has carried all along. Folds did not crash; they came out plausible and
systematically compressed, while pLDDT still read 87, because the confidence head predicts
error in the structure it is given rather than auditing its chemistry.

| | before | after | ideal |
|---|---|---|---|
| CA–CA mean | 3.27 Å | **3.67 Å** | 3.80 |
| CA–CA spread | 2.46–3.89 | **3.59–3.75** | — |
| helical CA | 5 | **16** | — |

No parity fixture could have caught it: those constants are not weights, and the sampler
draws its own noise so it has no fixture. It took folding something and measuring the bonds.

**0.1.3 — v2 would not load.** All three v2 packs died with
`Shapes (1,N,4,N,N) and (1,1,8,N,N) cannot be broadcast` after downloading up to 610 MB.
v2 is the only variant with `hidden_scale_up`, under which upstream recomputes pair heads
as `c_z / 32` — 8 for its 256-wide track. The trunk derived that correctly; the confidence
head hardcoded 4.

## Six packs

| id | size | cap | notes |
|---|---|---|---|
| `protenix-base-{int8,fp16,bf16}` | 225 / 469 / 418 MB | 700 | `int8` is the default choice |
| `protenix-v2-{int8,fp16,bf16}` | 299 / 610 / 561 MB | 250 | **mirror-sourced**, said in the `name` |

Generated from protenix-mlx's `WEIGHTS.md` rather than transcribed — writing it by hand, I
padded five 16-character prefixes from that file's display table into fake digests, which
would have failed only on a user's machine after a download. A test now asserts every
digest is real hex, unique, and free of the runs a padded one has.

`tiny` and `mini` are published upstream and **deliberately not registered**: 5 diffusion
steps does not converge the geometry (CA–CA 3.26 Å against base's 3.67), and a fold nobody
should trust is not worth its 11 seconds either. A test pins their absence.

v2 is capped at 250 because its memory is measured only at the short end and its pair track
is twice as wide, so base's curve *understates* it — the direction that gets a session
killed.

## Memory, measured

`ProtenixSizeGuard` is a separate type from `PredictSizeGuard`, not a parameterisation:
that guard's constants are Boltz-fitted and its docstring records three occasions where
extrapolating past its data ran optimistic. This one interpolates between measured points
and refuses beyond them.

| residues | 60 | 250 | 400 | 550 | 700 | 900 |
|---|---|---|---|---|---|---|
| peak (MiB) | 547 | 2279 | 3868 | 6303 | 8622 | 16513 |
| wall clock | — | 64 s | 150 s | 217 s | 377 s | **9054 s** |

The cap is **700**, and 900 is the reason it is not higher: 900 *fits* (16.1 GB against a
32 GB budget) and takes **2.5 hours** — 24× the wall clock for 1.3× the length, as the fold
stops being compute-bound and starts paging — for a structure whose own confidence is 26.4.
So the ceiling is "where the time stops being worth it", which is a judgement, not "where
the data runs out". The budget is 32 GB, applied as `min(budget, physical)`.

## What single-sequence folding is worth

Mean pLDDT on **complete domains**: villin 35aa 94.8, ubiquitin 76aa 67.5, barnase 110aa
58.9, lysozyme 129aa 37.0, ~26 from 400 up. Expected for an AF3-class model with no
alignment — the same pack scores villin 94.8 against poly-alanine 78.4 at equal length —
and it means this earns its time on small domains until #274.

## Also here

One unrelated commit, easy to drop: `RAYMOL_MCP_PORT`, because a dev build and an installed
RayMol cannot both bind 51737 and the second comes up with no MCP server at all. It is what
made the in-app verification above possible.

## Testing

299 predict tests, 274 Swift tests, 127 MSA tests. macOS builds. The iOS slice is untouched
— both new Swift files are inside `#if os(macOS)`, and its pre-existing `AppIcon` failure
reproduces identically on `master`.

## Deliberately not here

`metric_specs` is empty pending #308/#310: the runtime computes pLDDT, PAE, PDE and
resolved (verified to 2e-4), but a per-residue array needs the metric store to land in, and
declaring keys with nowhere to write them would put numbers in the schema that never
arrive. pLDDT still reaches the viewer through the B-factor column. Real MSAs (#274),
templates, ligands and nucleic acids remain unimplemented.

### DIFF
--- a/docs/predictors.md
+++ b/docs/predictors.md
@@ -23,16 +23,20 @@ it is shaped this way.
 |---|---|---|---|
 | `boltz2` | `boltz` | affine-int8, 507 MB | the default |
 | `boltz2-bf16` | `boltz` | dense bfloat16, 996 MB | same model, same runtime, unquantized |
-| `protenix-{tiny,mini,base,v2}-{int8,fp16,bf16}` | `protenix` | 84–582 MB | twelve packs, complexes, single sequence, confidence head |
+| `protenix-{base,v2}-{int8,fp16,bf16}` | `protenix` | 225–610 MB | six packs, complexes, single sequence, confidence head |
 
-`protenix-base-int8` is the default choice. tiny and mini are v0.5.0 models at 4 recycles /
-5 diffusion steps — seconds rather than minutes, and correspondingly rough. The dense
-precisions cost roughly twice the disk for no demonstrated accuracy, exactly as
-`boltz2-bf16` does. The `v2` packs are **mirror-sourced** — its official checkpoint has
-answered 403 since April 2026 — which their `name` says where a user picking a predictor
-will see it.
+`protenix-base-int8` is the default choice. The dense precisions cost roughly twice the
+disk for no demonstrated accuracy, exactly as `boltz2-bf16` does. The `v2` packs are
+**mirror-sourced** — its official checkpoint has answered 403 since April 2026 — which
+their `name` says where a user picking a predictor will see it.
 
-Twelve ids rather than one predictor with a `pack=` option, because that is what
+protenix-mlx also publishes `tiny` and `mini`, which are **deliberately not registered**:
+they are v0.5.0 models at 4 recycles / 5 diffusion steps, and five steps does not converge
+the geometry — CA-CA 3.26 Å against base's 3.67 and an ideal 3.80, loose enough that DSSP
+stops calling helices. They fold in 11 s against base's 35, but a fold nobody should trust
+is not worth 11 seconds either.
+
+Six ids rather than one predictor with a `pack=` option, because that is what
 `WeightBundle` already models: one bundle per predictor, pinned by digest. They are built
 from a table generated out of protenix-mlx's `WEIGHTS.md` rather than hand-copied — twelve
 transcribed digests is twelve chances to paste one that fails only on a user's machine,
--- a/docs/predictors.md
+++ b/docs/predictors.md
@@ -19,16 +19,39 @@ it is shaped this way.
 
 ## Shipped predictors
 
-| id | weights | notes |
-|---|---|---|
-| `boltz2` | affine-int8, 507 MB | the default |
-| `boltz2-bf16` | dense bfloat16, 996 MB | same model, same runtime, unquantized |
+| id | runtime | weights | notes |
+|---|---|---|---|
+| `boltz2` | `boltz` | affine-int8, 507 MB | the default |
+| `boltz2-bf16` | `boltz` | dense bfloat16, 996 MB | same model, same runtime, unquantized |
+| `protenix-base` | `protenix` | affine-int8, 214 MB | complexes, single sequence. **Not runnable yet** — no build carries the `protenix` runtime, see #309 |
 
-`boltz2-bf16` subclasses `Boltz2Predictor` and overrides nothing but `weight_bundle`.
-That works because `host.submit` sends `weights_dir` per job and the Swift runtime
-picks its matmul path from the artifact manifest — a dense pack declares no
-quantization block — so one runtime serves both with no host change. It needs
-boltz-mlx >= 0.2.0.
+## Runtimes
+
+A predictor's **runtime** is the Swift backend that runs it, named on the wire by
+`host.submit(..., runtime=)`. It is a separate axis from the predictor id: `boltz2` and
+`boltz2-bf16` are two predictors and one runtime, because `host.submit` sends `weights_dir`
+per job and the Swift side picks its matmul path from the artifact manifest — a dense pack
+declares no quantization block — so one runtime serves both.
+
+The distinction has to be on the wire because weights and featurizer are method-specific.
+Running one method's request on another's backend does not fail; it tokenizes with the
+wrong featurizer and returns a confident wrong structure. So:
+
+- `BoltzJobManager` implements `boltz` and **refuses** any other in `preflight`, before it
+  applies its size model — which is fitted to Boltz's measured peaks and would be
+  meaningless applied to another method's request even as a refusal.
+- `PyMOLBridge` advertises what is actually linked in `RAYMOL_PREDICT_RUNTIMES`, which is
+  what lets `check_available` refuse **before** a weight download rather than after.
+- `Request.runtime` is **optional** on the Swift side, absent meaning `boltz`. A
+  non-optional field would turn any Python/Swift version skew into "malformed prediction
+  request" instead of a refusal that names the missing runtime. The same reasoning applies
+  to `object_name` and `alignments`.
+
+That default is also the trap: a new predictor that forgets to pass `runtime=` is silently
+dispatched to Boltz. `predict_runtime.py` asserts every registered predictor names one.
+
+`boltz2-bf16` subclasses `Boltz2Predictor` and overrides nothing but `weight_bundle`, which
+is what "two predictors, one runtime" looks like in practice. It needs boltz-mlx >= 0.2.0.
 
 It is **not** an upgrade. On an M3 Pro at 117 tokens, recycling 3 / 200 steps, dense
 bfloat16 costs 2x the disk, +63% peak RSS (620 MiB -> 1012 MiB) and +22% wall clock
@@ -80,6 +103,13 @@ float16 is the closer one at identical size (`--precision float16` exports it).
 4. **Implement `check_available`** so the predictor disappears cleanly where it cannot run
    rather than failing mid-run. Platform, OS floor, host presence — not weight state, which the
    weight manager is allowed to fix by downloading.
+
+   If your method needs a Swift backend of its own, call `host.require_runtime` after
+   `host.require_available`: the two failures have different remedies ("you are headless"
+   versus "this build does not carry that backend"), and checking here is what refuses
+   BEFORE a multi-hundred-megabyte download rather than after it. A bulk
+   `predict_weights download=1` also consults this, so a predictor that cannot run is
+   reported but not fetched.
 
 5. **Implement `parse_spec` to reject, not repair.** If the backend silently ignores input it
    does not support, catching that is your job: check what it does with a ligand, a nucleic
--- a/docs/predictors.md
+++ b/docs/predictors.md
@@ -23,7 +23,21 @@ it is shaped this way.
 |---|---|---|---|
 | `boltz2` | `boltz` | affine-int8, 507 MB | the default |
 | `boltz2-bf16` | `boltz` | dense bfloat16, 996 MB | same model, same runtime, unquantized |
-| `protenix-base` | `protenix` | affine-int8, 214 MB | complexes, single sequence. **Not runnable yet** — no build carries the `protenix` runtime, see #309 |
+| `protenix-{tiny,mini,base,v2}-{int8,fp16,bf16}` | `protenix` | 84–582 MB | twelve packs, complexes, single sequence, confidence head |
+
+`protenix-base-int8` is the default choice. tiny and mini are v0.5.0 models at 4 recycles /
+5 diffusion steps — seconds rather than minutes, and correspondingly rough. The dense
+precisions cost roughly twice the disk for no demonstrated accuracy, exactly as
+`boltz2-bf16` does. The `v2` packs are **mirror-sourced** — its official checkpoint has
+answered 403 since April 2026 — which their `name` says where a user picking a predictor
+will see it.
+
+Twelve ids rather than one predictor with a `pack=` option, because that is what
+`WeightBundle` already models: one bundle per predictor, pinned by digest. They are built
+from a table generated out of protenix-mlx's `WEIGHTS.md` rather than hand-copied — twelve
+transcribed digests is twelve chances to paste one that fails only on a user's machine,
+after the download. Every id ends in a precision suffix so none is a prefix of another;
+see step 1.
 
 ## Runtimes
 
--- a/modules/pymol/predicting.py
+++ b/modules/pymol/predicting.py
@@ -1081,9 +1081,26 @@ SEE ALSO
     wait = not host.available() if int(async_) < 0 else not int(async_)
     out = {}
     for pid in ids:
-        bundle = registry.get(pid).weight_bundle
+        predictor_obj = registry.get(pid)
+        bundle = predictor_obj.weight_bundle
         if bundle is None:
             out[pid] = {'cached': True, 'path': None, 'bundle': None}
+            continue
+        # A BULK fetch downloads only what this build could actually run. Asking for one
+        # predictor BY NAME still downloads it -- that is someone pre-warming a cache
+        # deliberately, possibly for a build they do not have yet.
+        #
+        # Without this, "fetch everything" means every registered pack whether or not the
+        # host carries its runtime: `protenix-base` alone is 214 MB that no shipping build
+        # can use yet, pulled down by a command aimed at the predictors that do work.
+        if int(download) and not predictor and not _can_run(predictor_obj):
+            out[pid] = {'cached': bool(cache.is_cached(bundle)),
+                        'path': cache.path_for(bundle),
+                        'bundle': bundle.id}
+            if not int(quiet):
+                colorprinting.parrot(
+                    ' predict: %s cannot run in this build; not fetching its weights'
+                    % pid)
             continue
         if int(download) and not cache.is_cached(bundle):
             fetching.start(bundle, cache)
@@ -1096,6 +1113,21 @@ SEE ALSO
             colorprinting.parrot(' predict: %s weights cached=%s at %s' % (
                 pid, out[pid]['cached'], out[pid]['path']))
     return out
+
+
+def _can_run(predictor_obj):
+    """True if this predictor could run here, for deciding whether to pre-fetch.
+
+    Never raises: `check_available` is a predictor's own code, and a bulk pre-warm must
+    not be derailed by one method's availability check. An error means "cannot say", and a
+    bulk fetch treats that as "do not download" -- the cheap direction to be wrong in,
+    since naming the predictor explicitly still fetches it.
+    """
+    try:
+        predictor_obj.check_available()
+    except Exception:
+        return False
+    return True
 
 
 def _wait_for_fetch(bundle, quiet, _poll=0.1):
--- a/modules/pymol/predictors/__init__.py
+++ b/modules/pymol/predictors/__init__.py
@@ -12,10 +12,14 @@ def _register_builtins():
     """Register the shipped predictors. The only function that changes per predictor."""
     from .boltz2 import Boltz2Predictor
     from .boltz2_bf16 import Boltz2BF16Predictor
-    from .protenix import ProtenixBasePredictor
+    from .protenix import PREDICTORS as PROTENIX_PREDICTORS
     register(Boltz2Predictor(), replace=True)
     register(Boltz2BF16Predictor(), replace=True)
-    register(ProtenixBasePredictor(), replace=True)
+    # One id per published pack -- nine of them, variant x precision. A loop rather than
+    # nine lines because they differ only in which pack they name, and the list is
+    # generated from the digests protenix-mlx publishes; see protenix.py.
+    for predictor in PROTENIX_PREDICTORS:
+        register(predictor(), replace=True)
 
 
 _register_builtins()
--- a/modules/pymol/predictors/__init__.py
+++ b/modules/pymol/predictors/__init__.py
@@ -12,8 +12,10 @@ def _register_builtins():
     """Register the shipped predictors. The only function that changes per predictor."""
     from .boltz2 import Boltz2Predictor
     from .boltz2_bf16 import Boltz2BF16Predictor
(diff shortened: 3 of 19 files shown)


javierbq:
## All twelve packs are now registered, and v2 needed a fix to run at all

Pushed. Four variants × three precisions, each an id with its own digest-pinned bundle.

| variant | int8 | fp16 | bf16 | schedule | cap |
|---|---|---|---|---|---|
| tiny v0.5.0 | 87 MB | 173 MB | 152 MB | 4 recycles / 5 steps | 400 |
| mini v0.5.0 | 96 MB | 193 MB | 172 MB | 4 / 5 | 400 |
| base v1.0.0 | 225 MB | 469 MB | 418 MB | 10 / 200 | 400 |
| v2 *(mirror)* | 299 MB | 610 MB | 561 MB | 10 / 200 | **250** |

`protenix-base` → **`protenix-base-int8`**. Nothing has shipped, so the rename is free now
and permanent later: every id ends in a precision suffix, so none is a prefix of another
and `predict protenix-base-i<Tab>` completes to a runnable command rather than stopping
dead the way `boltz2` does for `boltz2-bf16`.

### v2 did not load — and it was my bug

All three v2 packs died on load with
`Shapes (1,N,4,N,N) and (1,1,8,N,N) cannot be broadcast`, **after** downloading 285–610 MB,
as an MLX fatal error rather than a refusal.

v2 is the only variant with `hidden_scale_up`, under which upstream recomputes
`no_heads_pair = c_z / c_hidden_pair_att` — so its 256-wide pair track needs 8
triangle-attention heads where every other variant's 128-wide track needs 4. The trunk had
always derived this correctly; the confidence head I wired earlier in this PR hardcoded
the 4. Fixed in protenix-mlx **0.1.3**, with a regression test. v2 int8 now scores pLDDT
90.2 on a short probe — the highest of any variant.

Found by smoke-testing all twelve packs instead of assuming the published ones work. Worth
doing: "available" has to mean "runs".

### The pack table is generated, not transcribed

From protenix-mlx's `WEIGHTS.md`, deliberately. Writing it by hand I padded five
16-character prefixes out of that file's display table into fake 64-character digests —
exactly the failure the comment above the table warns about, and one that fails only on a
user's machine, after the download. There is now a test asserting every digest is real
hex, unique across packs, and free of the repeated runs a padded one has.

### Two things measured rather than assumed

**Per-variant schedules.** tiny and mini are v0.5.0 models trained at 4 recycles / 5 steps;
running them at base's 10 / 200 would spend forty times the compute on a schedule they were
not trained for, and nothing would fail to tell you.

**Per-variant memory**, int8, MLX's own peak:

| | 60 res | 250 res | 400 res |
|---|---|---|---|
| tiny | 250 | 2083 | 3494 |
| mini | 298 | 2172 | 3625 |
| base | 547 | 2279 | 3868 |

They converge as the input grows — within 10% at 400 residues — because the N² pair
representation dwarfs the weights. So the cap is the same for all three rather than scaled
by parameter count. v2 is capped at 250 instead: its memory is measured only at the short
end and its pair track is twice as wide, so base's curve *understates* it, which is the
direction that gets a session jetsam-killed.

### Verified in the app

All twelve listed with their sizes, schedules and caps. `protenix-tiny-int8` downloaded
(87 MB in 6 s) and folded a 35-residue sequence in **10.9 s at 215 MiB**, against base's
34.7 s and 490 MiB — which is the point of shipping the small ones.

297 predict tests.

PR #326, opened 2026-08-19
MCP broker: reach RayMol even when it is closed

javierbq:
Fixes the MCP error you get when opening a Claude client while RayMol is closed.

Implements `docs/superpowers/specs/2026-08-19-raymol-mcp-broker-design.md`.

## The two defects

- **Claude Code** was registered as a pinned loopback URL (`--transport http http://127.0.0.1:<ephemeral>/mcp`), a snapshot of a port that changes every restart. Wrong whenever RayMol is closed, stale whenever it restarts.
- **Claude desktop app** registered `Bundle.main.executablePath`, so a Debug build run from a worktree wrote *its own* path into `claude_desktop_config.json` — a hard spawn error once that build dir is cleaned.

## The change

Both clients now register the same stdio broker (`RayMol --mcp-bridge`), which resolves a target from a per-instance registry and cold-launches the installed app on the first tool call.

- **Registry** — each running RayMol writes `~/Library/Application Support/RayMol/instances/<pid>.json` (mode 0600). The legacy `mcp.json` is still written and is still read as a fallback.
- **Resolution** — explicit `instance` argument → `RAYMOL_MCP_INSTANCE` → sticky session binding → registry scan. Two or more instances returns a listing and asks; it never guesses which window to drive.
- **Cold launch** — only `tools/call` launches. Opening a client answers `initialize`/`tools/list` locally, so it stops erroring without popping a window open.
- **Trust is unchanged** — a model-initiated launch still raises RayMol's Allow prompt.
- New `list_raymol_instances` tool; every tool gains an optional `instance` override.

**No login-item daemon.** Both clients spawn processes, so a resident background executable would add a login item, a restart story and a second artifact to notarize for no capability that gets used. The spec argues this explicitly, and the registry is the piece a daemon would need if a non-spawning client ever shows up.

## Four defects the end-to-end harness found that unit tests could not

1. `NSWorkspace` activated an unrelated dev build instead of launching the requested bundle — every RayMol shares `io.raymol.RayMol`, so the launch needs `createsNewApplicationInstance`.
2. The launch error was discarded, so a LaunchServices failure reported itself as "the MCP server didn't start" and sent the user to the wrong setting. Each outcome now names its own cause.
3. Cold launch persisted `raymol.mcp.enabled = true`, arming the MCP server for every future *manual* launch — a decision only the user gets to make.
4. A pre-registry RayMol never registers, so the broker cold-launched a **second** copy on top of the one in use. The legacy handoff is now tried first.

## Verification

`swiftui/tools/mcp_broker_e2e.sh` drives a suffixed dev build via `RAYMOL_MCP_INSTALLED_APP`, so it never overwrites the installed app.

- Rebased onto `master` (was 21 commits behind; only conflict was the generated `project.pbxproj`, resolved by regenerating with xcodegen — every MCP source file is byte-identical post-rebase).
- Unit suite on the rebased branch: **369 tests, 0 failures, 4 skipped** (33 new in `MCPBrokerTests`; the rest includes master's `PredictControllerTests`). The scheme builds `PyMOLViewer_macOS`, so the app target compiles against master's code too.
- Handshake launches nothing; `tools/call` cold-launches, the instance registers, and the reply is the Allow prompt.
- With a pre-registry RayMol running, a tool call reached **it** (real session state) and launched no second copy.
- Two instances: ambiguity listing names both keys; `instance: "RayMol-broker2"` executed on that instance while the other returned its own untrusted reply; an unknown key errors.
- After a cold launch with the preference deleted, `raymol.mcp.enabled` is still absent; stale registry entries from killed instances were pruned on the next scan.

## Not done here

The two broken configs on the dev machine are untouched — repairing them is a local change, and the full benefit only lands once a build containing this is installed.

### DIFF
--- a/docs/superpowers/plans/2026-08-19-raymol-mcp-broker.md
+++ b/docs/superpowers/plans/2026-08-19-raymol-mcp-broker.md
@@ -752,8 +752,14 @@ In `response(for:method:id:raw:)` (line 96), replace the opening `if let body = 
         let target = resolveTarget(requestedKey: nil)
         var instance: MCPInstance? = nil
         if case .bound(let i) = target { instance = i } else { instance = legacyHandoff() }
-        if let instance, let body = proxy(raw: raw, to: instance) {
-            return body.isEmpty ? nil : body   // empty = 202 notification ack
+        if let instance {
+            if let body = proxy(raw: raw, to: instance) {
+                return body.isEmpty ? nil : body   // empty = 202 notification ack
+            }
+            // Unreachable or 401 (a recycled pid, or a rotated token): the binding
+            // is stale, so drop it and let the next call re-resolve rather than
+            // retrying a port that is not ours.
+            setBound(nil)
         }
 ```
 
@@ -1354,7 +1360,7 @@ BODY
 
 **Spec coverage.** Registry file + fields → Task 4. Broker resolution order → Task 3, wired in Task 5. Ambiguity listing → Task 6. Instance keys with `#pid` collisions → Task 2. `instance` injection/stripping + `list_raymol_instances` → Task 6. Cold launch, defaults flip, 20 s poll, no launch lock → Task 7. Only-`tools/call`-launches → Task 7. Trust unchanged → no task sets `set_trusted`; asserted by the Task 9 Allow-prompt step. Error table → Tasks 6–7 (`ambiguous`, `unknownKey`, missing bundle, timeout) and Task 4 (`liveDirectoryInstances` prunes corrupt and dead entries). Migration → Task 8 plus Task 9 Step 5. Testing → Tasks 1–8 unit tests, Task 9 end-to-end.
 
-**Gap found and closed.** The spec lists "port returns 401 (stale token) → prune, rescan". Pid-liveness alone misses a pid that was recycled by an unrelated process. `liveDirectoryInstances` prunes dead pids only; the 401 case is left to `proxy` returning nil, which falls through to `legacyHandoff` and then to the offline reply. That is acceptable behavior but is NOT what the spec says. Task 5's `proxy` therefore must be treated as the 401 handler — the implementer should clear the sticky binding on a nil proxy result so the next call re-resolves. Add this to Task 5 Step 3: in `response`, on `proxy(...) == nil`, call `setBound(nil)` before falling through.
+**Gap found and closed.** The spec lists "port returns 401 (stale token) → prune, rescan". Pid-liveness alone misses a pid recycled by an unrelated process, so `liveDirectoryInstances` cannot be the whole answer. Fixed inline in Task 5 Step 3: a nil `proxy` result now clears the sticky binding, so the next call re-resolves instead of retrying a port that is no longer ours.
 
 **Placeholder scan.** No TBD/TODO. Every code step carries real code. Task 5 Step 1 honestly states it has no unit seam rather than inventing one.
 
--- a/docs/superpowers/plans/2026-08-19-raymol-mcp-broker.md
+++ b/docs/superpowers/plans/2026-08-19-raymol-mcp-broker.md
@@ -1,0 +1,1361 @@
+# RayMol MCP Broker Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Make RayMol's MCP server reachable from a Claude client even when RayMol is closed, by turning the existing `--mcp-bridge` stdio proxy into a broker that discovers running instances through a registry and cold-launches the installed app on the first tool call.
+
+**Architecture:** Each running RayMol writes a per-process JSON file into an instance registry directory. The broker (`RayMol --mcp-bridge`, a stdio process the Claude client spawns) reads that registry to resolve a target, launches `/Applications/RayMol.app` when nothing is running, and proxies JSON-RPC over loopback HTTP to the chosen instance. No resident daemon is introduced.
+
+**Tech Stack:** Swift 5 / AppKit / Foundation (macOS-only, inside `#if os(macOS) && !RAYMOL_MAS_RESTRICTED`), XCTest, XcodeGen, Python 3 (read-only reference to `modules/raymol_mcp/tools.py`).
+
+**Spec:** `docs/superpowers/specs/2026-08-19-raymol-mcp-broker-design.md`
+
+## Global Constraints
+
+- All new code is macOS-only and MUST sit inside `#if os(macOS) && !RAYMOL_MAS_RESTRICTED`, matching every file it touches. The Mac App Store build compiles MCP out entirely; a symbol leaking outside the guard breaks that build.
+- Registry directory: `~/Library/Application Support/RayMol/instances/`, one file per process named `<pid>.json`, mode `0600`.
+- Registry entry keys, exactly: `pid` (Int), `port` (Int), `token` (String), `appPath` (String), `name` (String), `installed` (Bool), `startedAt` (String, ISO-8601).
+- The canonical `~/Library/Application Support/RayMol/mcp.json` handoff MUST keep being written and removed exactly as it is today. The registry is additive.
+- Cold-launch target: `/Applications/RayMol.app`. Cold-launch timeout: 20 seconds. Existing 120-second proxy timeout is unchanged.
+- Only `tools/call` may cold-launch. `initialize`, `tools/list` and `ping` are answered locally when no instance is live.
+- Trust is unchanged: the broker never sets `set_trusted`. The Allow prompt must still appear on a model-initiated cold launch.
+- Instance key = bundle display name (`RayMol`, `RayMol-287`); on collision, `Name#<pid>`. A bare pid string is always accepted as a key.
+- Swift test files live in `swiftui/PyMOLViewerTests/`. After ADDING a file there you MUST run `xcodegen generate` in `swiftui/` before `xcodebuild`, because `PyMOLViewer.xcodeproj` is checked in.
+- Test command (run from `swiftui/`):
+  `xcodebuild test -project PyMOLViewer.xcodeproj -scheme UnitTests_macOS -destination 'platform=macOS' -skipPackagePluginValidation -skipMacroValidation -only-testing:PyMOLViewerTests/<Class> 2>&1 | tail -30`
+- Do NOT commit to `master`. Work stays on the current branch; a PR is opened at the end.
+
+---
+
+## File Structure
+
+| File | Responsibility |
+|---|---|
+| `swiftui/PyMOLViewer/Shared/MCPInstanceRegistry.swift` (create) | Pure model + pure resolution logic: the `MCPInstance` record, encode/decode, liveness filtering, key assignment, target selection. No I/O policy, no launching. Split out of `MCPBridge` so it can be unit-tested without a stdio loop. |
+| `swiftui/PyMOLViewer/Shared/MCPServerManager.swift` (modify) | Writes and removes this process's registry entry alongside the existing handoff; `connectClaudeCode` registers the stdio broker instead of a pinned HTTP URL. |
+| `swiftui/PyMOLViewer/Shared/MCPBridge.swift` (modify) | Broker behavior: resolve a target via `MCPInstanceRegistry`, cold-launch, sticky binding, `instance` argument injection/stripping, `list_raymol_instances`. |
+| `swiftui/PyMOLViewer/Shared/MCPDesktopInstaller.swift` (modify) | `bridgeCommand()` prefers the installed app over the running bundle. |
+| `swiftui/PyMOLViewerTests/MCPBrokerTests.swift` (create) | Unit tests for every pure function above. |
+
+Tasks 1–3 build the pure core bottom-up; Task 4 wires the registry writer; Tasks 5–7 change broker behavior; Task 8 fixes registration paths; Task 9 is end-to-end verification and config repair.
+
+---
+
+### Task 1: Instance record and registry decoding
+
+**Files:**
+- Create: `swiftui/PyMOLViewer/Shared/MCPInstanceRegistry.swift`
+- Create: `swiftui/PyMOLViewerTests/MCPBrokerTests.swift`
+
+**Interfaces:**
+- Consumes: nothing.
+- Produces:
+  - `struct MCPInstance: Codable, Equatable` with `let pid: Int`, `let port: Int`, `let token: String`, `let appPath: String`, `let name: String`, `let installed: Bool`, `let startedAt: String`
+  - `enum MCPInstanceRegistry` with `static func decode(_ data: Data) -> MCPInstance?`
+  - `static func decodeAll(_ blobs: [Data]) -> [MCPInstance]`
+
+- [ ] **Step 1: Write the failing test**
+
+Create `swiftui/PyMOLViewerTests/MCPBrokerTests.swift`:
+
+```swift
+import XCTest
+@testable import RayMol
+
+/// Covers the pure core of the MCP broker: how a client decides WHICH RayMol
+/// it is talking to.
+///
+/// This matters because the broker is spawned by a Claude client with no user
+/// present to correct it. A stale registry entry that survives filtering means
+/// tool calls are proxied to a dead port; a wrong disambiguation means Claude
+/// silently drives a different window than the one the user is looking at.
+final class MCPBrokerTests: XCTestCase {
+
+    // MARK: - decode
+
+    private func blob(pid: Int = 4412, port: Int = 51737, name: String = "RayMol",
+                      installed: Bool = true) -> Data {
+        Data("""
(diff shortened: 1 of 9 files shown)


javierbq:
## Rebased onto master + how to test this on another Mac

Rebased onto `master` (`4f9f3b091`, the #329 predict-superpose fix) — clean, no conflicts. The only file both sides touched was the generated `project.pbxproj`, regenerated with `xcodegen`.

### Verification after the rebase

| Check | Result |
|---|---|
| `xcodebuild test -scheme UnitTests_macOS` | **401 tests, 0 failures**, 4 skipped (the inference ones) |
| iOS device build (`generic/platform=iOS`) | **BUILD SUCCEEDED** — no platform-symbol leak from the new shared file |
| Mac App Store variant (`RAYMOL_MAS_RESTRICTED`) | **BUILD SUCCEEDED** — MCP compiles out cleanly |

---

## How much of this is actually tested

Being explicit, because the gap matters for what you should poke at by hand.

### Verified by running it

- **32 unit tests** in `MCPBrokerTests` covering the pure core: registry decode (including corrupt/truncated entries), dead-pid pruning, key assignment with the `Name#pid` collision case, `match` refusing an ambiguous bare name, the full resolution precedence (explicit → env → sticky → scan), and each cold-launch failure message.
- **Cold launch, end to end.** With the registry empty, a `tools/call` launched the app, it registered, and the reply was the Allow prompt.
- **Handshake launches nothing.** `initialize` + `tools/list` with nothing running left zero processes started.
- **Migration path.** With a pre-registry RayMol (1.9.1) running, a tool call reached *it* — real session state back — and launched no second copy.
- **Ambiguity + routing.** Two registered instances produced the listing with both keys; `instance: "RayMol-broker2"` executed Python on that instance while the other returned its own untrusted reply; an unknown key errored. Routing was made falsifiable by giving only one instance `RAYMOL_MCP_AUTOTRUST=1`.
- **No persisted preference.** With `raymol.mcp.enabled` deleted, a cold launch registered a live instance and the preference was still absent afterwards.
- **Stale pruning.** Entries left by `kill`ed instances were gone on the next scan.

### NOT tested — worth your attention

1. **Neither real client was used.** All protocol traffic was raw JSON-RPC piped into `--mcp-bridge`. How the Claude desktop app and Claude Code actually spawn it, and their handshake quirks, are unverified. **This is the main thing to test.**
2. **Cold launch never ran against a real `/Applications/RayMol.app`** — always through the `RAYMOL_MCP_INSTALLED_APP` override on a suffixed dev build, so as not to overwrite the installed app.
3. **The Allow prompt was never clicked.** "Click Allow, retry, it works" is inferred; trusted execution was proven separately via auto-trust.
4. **No clean-machine run** — never tested with no prior `mcp.json` and no token in defaults.
5. **Two brokers racing** to cold-launch is reasoned about, not exercised.
6. **Session continuity** when `initialize` lands on a legacy instance and a later call re-resolves elsewhere.
7. **iOS simulator build** could not link in my worktree (no prebuilt simulator `libpymol_core.a`); the device build covers the compile.

---

## Test recipe for another Mac

### 1. Build

```bash
git fetch origin claude/raymol-mcp-companion-477cf0 && git checkout claude/raymol-mcp-companion-477cf0
cd swiftui && ./build_macos.sh && xcodegen generate
xcodebuild build -project PyMOLViewer.xcodeproj -scheme PyMOLViewer_macOS -destination 'platform=macOS' -derivedDataPath build_mac_dd -skipPackagePluginValidation -skipMacroValidation
```

The C++ core must be built **before** `xcodebuild` — `xcodebuild` alone silently links a stale `libpymol_core.a`.

### 2. Make a suffixed, re-signed dev bundle

Per `CLAUDE.md`, never leave a dev build named `RayMol.app`. Renaming invalidates the ad-hoc signature, hence the re-sign:

```bash
cd swiftui && DST=/tmp/RayMol-broker.app && rm -rf "$DST" && cp -R build_mac_dd/Build/Products/Debug/RayMol.app "$DST" && mv "$DST/Contents/MacOS/RayMol" "$DST/Contents/MacOS/RayMol-broker" && for k in CFBundleExecutable CFBundleName CFBundleDisplayName; do plutil -replace $k -string "RayMol-broker" "$DST/Contents/Info.plist"; done && codesign --force --deep --sign - "$DST"
```

### 3. Automated harness

```bash
./swiftui/tools/mcp_broker_e2e.sh /tmp/RayMol-broker.app
```

Expect: `OK: handshake did not launch the app`, then `OK: RayMol-broker is running` with 1 registry entry. Prefix `RAYMOL_MCP_PORT=0` **only** if another RayMol already holds 51737 — otherwise the launched app fails to bind, never registers, and you get the 20s timeout.

### 4. The important one: a real Claude client

```bash
claude mcp remove raymol --scope user
claude mcp add raymol --scope user -e RAYMOL_MCP_INSTALLED_APP=/tmp/RayMol-broker.app -- /tmp/RayMol-broker.app/Contents/MacOS/RayMol-broker --mcp-bridge
```

Then, **with RayMol closed**, restart Claude Code and check in order:

1. `/mcp` shows `raymol` connected, **not** errored — this is the original bug.
2. Ask it to load a structure ("show me 1ubq in RayMol"). RayMol should open on its own, and RayMol should raise **Allow**.
3. Click Allow, ask again — it should now work. *(Step 3 is the specific thing I could not verify.)*
4. Quit RayMol mid-conversation, then ask for something else — it should launch again rather than error.

For the desktop app, the equivalent entry is `{"command": "/tmp/RayMol-broker.app/Contents/MacOS/RayMol-broker", "args": ["--mcp-bridge"], "env": {"RAYMOL_MCP_INSTALLED_APP": "/tmp/RayMol-broker.app"}}` in `~/Library/Application Support/Claude/claude_desktop_config.json`. Quit and reopen Claude afterwards.

### 5. Multi-instance behavior

```bash
DST=/tmp/RayMol-broker2.app && rm -rf "$DST" && cp -R /tmp/RayMol-broker.app "$DST" && mv "$DST/Contents/MacOS/RayMol-broker" "$DST/Contents/MacOS/RayMol-broker2" && for k in CFBundleExecutable CFBundleName CFBundleDisplayName; do plutil -replace $k -string "RayMol-broker2" "$DST/Contents/Info.plist"; done && codesign --force --deep --sign - "$DST"
RAYMOL_MCP_ENABLE=1 RAYMOL_MCP_PORT=0 /tmp/RayMol-broker.app/Contents/MacOS/RayMol-broker &
RAYMOL_MCP_ENABLE=1 RAYMOL_MCP_PORT=0 /tmp/RayMol-broker2.app/Contents/MacOS/RayMol-broker2 &
```

With both up, a tool call should return the "More than one RayMol is running" listing rather than picking one; `list_raymol_instances` should show both; and passing `instance: "RayMol-broker2"` should hit that one specifically.

### Cleanup

```bash
pkill -x RayMol-broker; pkill -x RayMol-broker2; rm -rf /tmp/RayMol-broker.app /tmp/RayMol-broker2.app
rm -f ~/Library/Application\ Support/RayMol/instances/*.json
claude mcp remove raymol --scope user
```

### One caveat about testing on a second machine

Every RayMol build shares `CFBundleIdentifier = io.raymol.RayMol`. If that Mac has other RayMol builds running, the broker will (correctly) report ambiguity instead of picking one — and if one of them holds port 51737, a newly launched instance can't bind and won't register. Quit stray builds first, or pass `RAYMOL_MCP_PORT=0`.

javierbq:
## Functional test on a real Mac — PR #326

Tested `claude/raymol-mcp-companion-477cf0` @ `5000fd4f2` on macOS 26 (Darwin 25.5.0), Xcode 26.6 /
Swift 6.3.3, M-series. Built the C++ core first (`bash swiftui/build_macos.sh` → `libpymol_core.a`),
then `xcodegen generate` + `xcodebuild` (`** BUILD SUCCEEDED **`), then a suffixed re-signed bundle
`/tmp/RayMol-broker.app` (`CFBundleName=RayMol-broker`). `/Applications/RayMol.app` was never touched.

**Headline: the original bug is fixed, and I found one new blocking bug in the cold-launch path —
the "Allow" prompt is never actually raised, so step 3 of Test 2 cannot happen as written.** Fix +
4 tests pushed to this branch.

---

### Baseline — the bug this PR fixes, reproduced first

Before changing anything, the pre-existing user-scope entry was the stale pinned URL:

```
raymol: http://127.0.0.1:51737/mcp (HTTP) - ✘ Failed to connect — ConnectionRefused
```

with `~/.claude.json` holding `{"type":"http","url":"http://127.0.0.1:51737/mcp", ...}` and a
matching token in a leftover `~/Library/Application Support/RayMol/mcp-58647.json` (pid 58647 long
dead). So the reported symptom was real on this machine.

After registering with the arg vector the installer actually emits
(`MCPServerManager.claudeCodeAddArgs`, i.e. `claude mcp add raymol --scope user -- <bin> --mcp-bridge`,
**no** `-e` in the product path):

```
raymol: /tmp/RayMol-broker.app/Contents/MacOS/RayMol-broker --mcp-bridge - ✔ Connected
```

and `claude mcp list`'s health check launched nothing (`pgrep -x RayMol-broker` empty afterwards).
That is the original bug fixed, verified through the real CLI rather than by piping JSON.

### Test 1 — automated harness

`./swiftui/tools/mcp_broker_e2e.sh /tmp/RayMol-broker.app` — passes exactly as documented:
`OK: handshake did not launch the app`, `OK: RayMol-broker is running`, `registry now: 1 entries`.

### Test 2 — real Claude Code client

I could not restart *this* Claude Code session to pick up a user-scope entry (it is the session doing
the testing), so I drove a genuine Claude Code MCP client out-of-process with
`claude -p --mcp-config <file> --strict-mcp-config`. That is the real client implementation spawning
the real stdio broker and doing the real handshake — only the transcript is headless.

**Step 1 — `/mcp` shows connected, not errored: PASS.** The client listed all six tools
(`run_pymol_command`, `run_python`, `get_session_state`, `capture_viewport`, `search_pdb`,
`list_raymol_instances`) with RayMol closed, and `pgrep` confirmed **nothing was launched** by the
handshake. `claude mcp list` independently reports `✔ Connected` (above).

**Step 2 — ask for a structure, RayMol opens by itself: PASS on the launch, FAIL on the prompt.**
"Show me 1ubq in RayMol" cold-launched the app from the real client (new pid, registry entry written,
port 51737). Verbatim tool results the client reported:

```
list_raymol_instances -> No RayMol is running.
run_pymol_command {"command":"fetch 1ubq, async=0"} ->
  RayMol is waiting for the user to approve this connection. Ask the user to click Allow in RayMol, then retry.
```

The cold launch works. **But no Allow prompt is raised** — see the bug section below.

**Step 3 — click Allow, retry, it works: BLOCKED (this was the unverified item, and it does not
hold).** There is no prompt to click on the cold-launch path. Retrying returns the same refusal
indefinitely — three retries in one broker session, all identical:

```
id 2 -> RayMol is waiting for the user to approve this connection. Ask the user to click Allow in RayMol, then retry.
id 3 -> RayMol is waiting for the user to approve this connection. Ask the user to click Allow in RayMol, then retry.
id 4 -> RayMol is waiting for the user to approve this connection. Ask the user to click Allow in RayMol, then retry.
```

**Step 4 — quit RayMol mid-conversation, ask again: PASS.** With a live bound instance (pid 14312),
`SIGTERM`, then another tool call on the *same* broker session: the stale registry entry was pruned
and a fresh app was launched (pid 14457) rather than erroring. Relaunch works; it then lands in the
same approval dead end as step 2.

> Note for anyone repeating step 4: `pgrep -x RayMol-broker` matches **both** the GUI app and the
> `--mcp-bridge` process, because the broker is the same executable. My first attempt killed the
> broker itself (`BrokenPipeError`). Kill the pid from the registry entry instead.

**Desktop app — NOT VERIFIED, and I did not attempt it.** This session runs inside
`/Applications/Claude.app` (ancestor pid 82179), so quitting and reopening Claude to load
`claude_desktop_config.json` would terminate the session doing the test. I left that config file
untouched rather than write an entry pointing at a `/tmp` bundle I then delete. The broker side is
transport-identical (same stdio command, same handshake, and the client-spawn path is now proven with
a real MCP client), but **the desktop app's own spawn/handshake remains unverified** — happy to do it
in a throwaway session if you want it covered.

### Test 3 — multi-instance: PASS

Second bundle `/tmp/RayMol-broker2.app`, both launched with `RAYMOL_MCP_PORT=0` (ports 52873 / 52875).

- Tool call with two running returns the listing rather than picking one:
  ```
  More than one RayMol is running. Ask the user which one to drive, then retry with the instance argument set:
    • RayMol-broker (dev build) — /tmp/RayMol-broker.app
    • RayMol-broker2 (dev build) — /tmp/RayMol-broker2.app
  ```
- `list_raymol_instances` shows both with pid/port/path.
- `instance: "RayMol-broker2"` routed correctly — verified by querying each port directly afterwards:
  port 52875 had `['1crn']`, port 52873 had `[]`.
- `instance: "RayMol-nope"` → `No running RayMol named "RayMol-nope". Call list_raymol_instances.`

---

## Bug found: the cold-launch path tells the user to click Allow but never raises the prompt

**Cause.** RayMol's server emits the connect event that raises the approval alert **only** on
`initialize` (`server.py`: `if is_init: ... if not _trusted: events.client_connected(session_id)` →
`handleFeedbackEvent("connect")` → `pendingApproval = true` → the alert). But when RayMol is down the
broker answers `initialize` locally (correctly — that is what keeps a client from opening a window),
and on the cold-launch branch it then forwards **only the `tools/call`**. So the freshly launched
server's first-ever request is the tool call: `_trusted` is false, it returns "click Allow in
RayMol", and no prompt was ever raised. Every retry repeats it. Same dead end after a step-4 relaunch,
and on the manual "open RayMol after the client already handshook" path.

**How I proved it.** Accessibility and screen recording are both denied to this session, so I used
`CGWindowListCopyWindowInfo` (needs neither) to count the app's windows. Within one instance:

| event | windows |
|---|---|
| cold-launched, 3 refused tool calls, +6s settle | 7 (no 260×238) |
| one more bare `tools/call` | 7 (no 260×238) |
| one `initialize` (curl, same port+token) | **260×238 layer-0 window appears** |

Reproduced on two independent instances. Window titles are redacted without screen-recording
permission, so I identified the alert by correlation (it appears only in response to `initialize`,
which is the only trigger for `client_connected`, and its geometry is stable), not by reading its text.

**Fix** (`swiftui/PyMOLViewer/Shared/MCPBridge.swift`): before forwarding to an instance, if we hold
no session, send the `initialize` ourselves — which is exactly what makes the server raise the prompt.

- `brokerInitializeRequest()` / `needsSessionHandshake(sessionId:method:)` — pure, unit-testable.
- `ensureSession(with:method:)` called on the common forwarding path, so it covers cold launch,
  step-4 relaunch, and the legacy-handoff branch alike.
- Also: `setBound` now clears the session id when the bound pid changes. A session id belongs to one
  server process; replaying it at a relaunched RayMol names a session that server never issued.

**After the fix**, the same cold launch through the broker leaves the 260×238 alert on screen
immediately, so "click Allow, then retry" is finally a real path.

Pushed as `20d8a868` on this branch. Unit tests: `** TEST SUCCEEDED **`, **405 tests, 0 failures**
(401 → 405, exactly the 4 added). Re-ran Test 1 and the trusted-instance end-to-end after the fix —
no regression. Final confirmation with a real Claude Code client and RayMol closed: cold launch fired
**and** the 260×238 alert was on screen, so the "click Allow" message now matches reality.

**Tests added** (`MCPBrokerTests.swift`, 4) — `testForwardingWithoutASessionHandshakesFirst` is the
one that pins the actual defect (step 2/3 caught it); the others pin that an established session is
not re-handshaken, that a client `initialize` is not double-wrapped, and that what we send really is
an `initialize` with an id.

**Not verified:** I could not physically click Allow (no accessibility/GUI control in this session), so
the post-approval half was verified via the trusted-instance equivalent (`RAYMOL_MCP_AUTOTRUST=1`):
through the broker, `fetch 1ubq` → `ok` and `get_session_state` → `1ubq`, 660 atoms; and after the fix
`fetch 1crn` → `ok`, 327 atoms. The prompt-raising defect above is what I proved directly.

## Smaller observations

- **One transient auto-start failure, not reproducible.** On one launch the app came up (GUI, 7
  windows, LaunchServices-registered) but its MCP server never started — no port, no registry entry,
  after 2 minutes. Four later launches were fine, including 2/2 with `RAYMOL_MCP_PORT=0`, so the port
  override is not the cause. Suspect the `autoStartIfEnabled` retry cap: 40 × 0.25s = **10s** for
  `engine.isReady`, which a cold Debug first launch can plausibly exceed. Worth a longer cap or a
  log line, but I could not reproduce it — flagging, not claiming.
- `swiftui/build_macos.sh` is checked in mode `100644`, so the documented `./build_macos.sh` fails
  with permission denied; needs `bash build_macos.sh`. **Pre-existing on master**, not this PR.
- Every instance shares one token (persisted `raymol.mcp.token` in UserDefaults) — the same value as
  the 3-day-old leftover handoff file. **Pre-existing on master**, out of scope here; noting only
  because the broker now hands that one token to several instances at once. The trust gate, not the
  token, is what actually protects tool calls.

## Cleanup done

Stray builds from other worktrees (`RayMol-PR310/PR313/315-319`) were left running untouched; they
predate the registry so they never registered and never affected disambiguation. `/tmp/RayMol-broker*.app`
removed, registry cleared, `raymol` user-scope MCP entry removed, `~/.claude.json` backed up before
any change.

PR #328, opened 2026-08-20
feat(ios): port the Predict tool to iPhone (Boltz int8, on-device)

javierbq:
Ports structure prediction (`cmd.predict` / the Predict tool) to iOS so it can be tested
on a physical iPhone. Boltz int8 only — Protenix stays macOS-only by decision.

**Draft: the code is complete and both slices build, but the on-device run has not
happened yet** (the test phone was locked when the branch was pushed). The one constant
that must not ship un-measured is called out below.

## Why it did not exist on iOS

Three independent gates, all of which had to go:

1. Eleven `#if os(macOS)` blocks across `PredictBar`/`PredictController`/`BoltzJobManager`
   /`BoltzRuntime`/`PredictSizeGuard` and the `PREDICT:` feedback dispatch in
   `PyMOLEngine` — which IS the invocation path, since there is no Python→Swift call.
2. `platforms: [macOS]` on the `BoltzMLX` package, so the runtime was never linked into
   the iOS slice at all.
3. `RAYMOL_PREDICT_HOST` exported only under `TARGET_OS_OSX`, so
   `predictors/host.py:available()` was False and every predictor correctly refused.

## What changed

**Link and reach.** `BoltzMLX` loses its platform filter and links on both platforms,
Simulator included — that is the only way the iOS slice keeps *compiling* this code, and
no CI job builds iOS. `PyMOLBridge.mm` exports the host variables for iOS on real hardware
only, advertising `boltz` alone: Protenix peaks near 3.9 GB at 400 residues, so
`ProtenixJobManager`/`ProtenixSizeGuard` stay behind `#if os(macOS)` and the Python side
must not be told otherwise, or `check_available` would green-light a predictor whose Swift
half is not in the binary — after a weight download.

`PredictAvailability` is the runtime gate: iOS 18+, and **never the Simulator**, where
MLX's Metal-backed allocator builds heaps `MTLSimDevice` rejects outright. `PyMOLBridge.mm`
applies the same rule, so Python refuses in step with the UI rather than offering a form
whose every Run reports the predictor unavailable.

**An honest iOS memory guard — the substance of the PR.**

`PredictSizeGuard.availableBytes` was `ProcessInfo.physicalMemory`. On an iPhone 15 Pro
that reports 8 GB where the app may use a fraction of it, so it would have licensed runs
at roughly 3× the real budget — and the failure mode is a jetsam SIGKILL no Swift handler
can catch, which takes the unsaved session with it. It is now `os_proc_available_memory()`,
exactly as `DesignSizeGuard` already does.

**Reusing the Mac's fitted constants was tried, and it is wrong — not conservative,
unusable.** At the 117-token anchor the Mac curve estimates 2.25 GB; against a realistic
phone budget near 2.5 GB that is 86%, which this guard *refuses*. The Mac fit would have
made the 100–150-residue fold this port exists for impossible on the device it was ported
to. A guard that refuses everything is not a safe guard.

So iOS gets its own fit, anchored on the only iOS peak anyone has recorded (iPhone 15 Pro,
117 tokens / 899 atoms, ~1.4 GB `phys_footprint`, from boltz-mlx's `DEVICE_BENCHMARK.md`):
the Mac curve's shape, intercept set independently to 700 MiB as an envelope over the
529 MB int8 pack, linear and quadratic terms scaled together — which lands both at exactly
half their Mac values, consistent with iOS running under a 64 MiB MLX cache rather than
256 MB. Result at the anchor: **1.75 GB against a measured 1.4 GB, a 19% reserve.**

Two things the fit deliberately does not claim, because it is **one point** and this type's
doc records three separate optimistic failures, every one of them an extrapolation:

- `bytesPerTokenMSARow` is **not** rescaled. No MSA fold has been run on device, and
  halving it "to match" would be inventing a measurement.
- `iOSMaximumTokens` is **117** — the measured size. Not the 256 the runtime's phone
  preset allows, and not the 150 this port targets. **This is the number that must be
  raised only alongside a real run, and doing so is the remaining work on this PR.**

Also: `BoltzRuntime` asks for a 64 MB MLX cache on iOS — below Design's 96 MB, so min-wins
arbitration pulls Design's effective ceiling down too (intended; a smaller cache can only
churn, never fail, and it is called out in the doc rather than left to be discovered). It
installs `MLX.Memory.memoryLimit` at the app's whole jetsam ceiling, sized to the device
rather than the run so it cannot clamp Design, which converts an overshoot from an
uncatchable SIGKILL into a catchable MLX error. `BoltzJobManager` uses the phone
`BoltzInputLimits` on iOS, written out explicitly rather than obtained by omitting
arguments, so an upstream change to those defaults is a visible diff.

**iPhone UI.** `PredictCompactPanel`: two docked rows (status, then input · object ·
model · Run) plus a settings sheet, the same split `DesignCompactPanel` makes. `PredictBar`
is not reused — it is a ~700 pt macOS form and uses `.toggleStyle(.checkbox)`, which does
not exist on iOS. Wired into all four iOS layouts alongside `designModeBar`.

**Backgrounding.** `project.yml` recorded that no run above ~115 tokens had ever completed
on device because iOS suspended the app mid-run, and nothing in RayMol set
`isIdleTimerDisabled`. `PredictScreenAwake` now holds the display on while a fold *or* the
529 MB weight download is in flight — attached at the ContentView root and to both feeds,
so it survives the user leaving the Predict tool while the job runs. It cannot make the app
background-safe (no background mode grants Metal compute), so a banner says the one thing
that loses the run.

## Not enabled: Increased Memory Limit

`PyMOLViewer/RayMolIOS.entitlements` requests
`com.apple.developer.kernel.increased-memory-limit`, which would roughly double the
foldable size — but it is **commented out** in `project.yml`, because it does not build
until "Increased Memory Limit" is enabled for the `io.raymol.RayMol` App ID in the
developer portal. That is an account change, so it is left as a one-line opt-in with the
exact steps in the comment.

## Verification

- **Both slices compiled by hand** (no CI builds iOS, and this boundary has broken in both
  directions before): `PyMOLViewer_iOS` for device and `PyMOLViewer_macOS` both
  `BUILD SUCCEEDED`.
- **358 macOS unit tests pass, 0 failures**, 21 of them new. The iOS rules are pure
  functions taking platform/budget as parameters specifically so the macOS test host can
  exercise them — a rule behind `#if os(iOS)` is a rule CI never runs. Includes a
  regression test pinning that the Mac fit would have refused the target workload, so a
  future "simplification" that deletes the iOS branch fails in CI rather than on a phone.
- **Bundle checked**: `BoltzMLX` symbols and the MLX metallib are in the iOS product, and
  `modules/pymol/predictors/*` + `msa.py` ship in the app.

## Outstanding

- [ ] Fold a ~100-residue single chain on the iPhone 15 Pro; record wall time and peak
      `phys_footprint`.
- [ ] Measure `os_proc_available_memory()` on device and confirm the 2.5 GB budget the fit
      was reasoned against.
- [ ] Raise `iOSMaximumTokens` to the largest size that actually completes — **or leave it
      at 117.** Never raise it alongside an extrapolation.

### DIFF
--- a/modules/pymol/appkit_predict.py
+++ b/modules/pymol/appkit_predict.py
@@ -16,7 +16,25 @@ from pymol import cmd
 
 
 def _predictors():
-    """[{'id': str, 'msa': bool}] for every registered predictor, sorted by id.
+    """[{'id': str, 'msa': bool}] for every registered predictor that can ACTUALLY RUN
+    on this host, sorted by id.
+
+    Filtered by each predictor's own `check_available()`, not merely by what is
+    registered. The registry is the same on every platform -- it is built by
+    `_register_builtins()` with no per-OS branches -- so an unfiltered list offers
+    methods the host cannot run: on iOS the app advertises the `boltz` runtime alone
+    (see PyMOLBridge.mm's RAYMOL_PREDICT_RUNTIMES), which makes all six Protenix packs
+    refuse at submit time. Listing them anyway turns a clear "not on this platform"
+    into a menu entry that fails after the user has typed a sequence and hit Run.
+
+    An empty list is a meaningful answer, not a bug to paper over: under headless
+    `pymol -c` nothing consumes the PREDICT: marker, so every predictor is genuinely
+    unavailable and the bar must offer nothing rather than something that would hang.
+
+    Never raises: `check_available` is a predictor's own code, and one bad method must
+    not empty the whole menu. A predictor that throws anything other than a refusal is
+    treated as unavailable -- the conservative direction, since the alternative is
+    offering a method whose availability we could not establish.
 
     `msa` is the method's own `supports_msa`: the bar disables the MSA controls for
     a method (e.g. protenix) that would refuse an alignment.
@@ -26,6 +44,13 @@ def _predictors():
     for pid in registry.available():
         try:
             p = registry.get(pid)
+        except Exception:
+            continue
+        try:
+            p.check_available()
+        except Exception:
+            continue
+        try:
             supports = bool(getattr(p, 'supports_msa', False))
         except Exception:
             supports = False
--- a/modules/pymol/predictors/boltz2_bf16.py
+++ b/modules/pymol/predictors/boltz2_bf16.py
@@ -25,7 +25,10 @@ would be a strictly closer one at identical size. Both are exportable
 (`boltz-mlx export-model --precision`); bfloat16 is offered because it is the
 width Boltz was trained in.
 """
+import sys
+
 from .boltz2 import Boltz2Predictor
+from .errors import PredictorUnavailable
 from .weights import WeightBundle
 
 
@@ -47,3 +50,37 @@ class Boltz2BF16Predictor(Boltz2Predictor):
         size=1_044_050_191,
         members=('config.json', 'manifest.json', 'model.safetensors'),
     )
+
+    def check_available(self):
+        """Available wherever Boltz is -- EXCEPT iOS.
+
+        Not a capability limit: the bf16 runtime is the same Swift `boltz` runtime the
+        int8 pack uses, so this method would load and run on a phone. It is refused
+        because the iOS memory guard cannot honestly admit it.
+
+        `PredictSizeGuard.decide` takes tokens, MSA depth and available bytes -- it does
+        NOT take a predictor, so there is one fitted curve for all Boltz packs, and that
+        curve was fitted to MEASURED int8 peak `phys_footprint` on an iPhone 15 Pro. The
+        dense pack needs materially more than int8 at the same token count (bigger
+        resident weights: 1.04 GB against 529 MB, plus wider activations). Feeding it
+        through the int8 curve therefore produces an estimate BELOW the real cost, which
+        is the one direction this guard must never err in: the failure mode is a jetsam
+        SIGKILL that no Swift handler can catch and that takes the unsaved session with
+        it. The guard's own doc comment records it having been wrong optimistically twice.
+
+        The rule that follows from that is "never let a fit sit below a measurement", and
+        no bf16 run has been measured on device. So the pack is refused rather than
+        guessed at. Lifting this needs a device sweep of bf16 peak footprints and a
+        predictor-aware `decide`, not a scaled-up int8 number.
+
+        macOS is unaffected: `availableBytes` there is real physical memory with a desktop
+        headroom budget, and the desktop has no jetsam.
+        """
+        super().check_available()
+        if sys.platform == 'ios':
+            raise PredictorUnavailable(
+                '%s is not available on iOS: the dense bfloat16 pack needs more memory '
+                'than the int8 pack the on-device size guard is calibrated against, and '
+                'no bfloat16 run has been measured on device. Use boltz2 (int8).'
+                % self.id
+            )
--- a/modules/pymol/predictors/weights.py
+++ b/modules/pymol/predictors/weights.py
@@ -135,8 +135,24 @@ class WeightCache:
     def default_root(platform=None, home=None):
         platform = sys.platform if platform is None else platform
         home = os.path.expanduser('~') if home is None else home
-        if platform == 'darwin':
+        if platform in ('darwin', 'ios'):
             # Under the App Sandbox this same path resolves inside the container.
+            #
+            # 'ios' is NOT cosmetic. CPython reports sys.platform == 'ios' there, not
+            # 'darwin', so an iPhone used to fall through to the ~/.raymol branch below --
+            # and on iOS $HOME is the app's DATA CONTAINER ROOT, which is readable but not
+            # writable. The download failed at the first mkdir with
+            #   [Errno 1] Operation not permitted:
+            #   /var/mobile/Containers/Data/Application/<uuid>/.raymol
+            # after the user had already asked for a 529 MB fetch. Only Documents/,
+            # Library/ and tmp/ are writable inside the container, so the same
+            # Library/Application Support path the sandboxed Mac build uses is both the
+            # correct location here and already covered by this branch.
+            #
+            # Library/Application Support rather than Documents/ on purpose: weights are
+            # a reproducible cache, not user data, and Documents/ is surfaced in the Files
+            # app (UIFileSharingEnabled) where half a gigabyte of model shards would sit
+            # among the user's own structures inviting deletion.
             return os.path.join(home, 'Library', 'Application Support',
                                 'RayMol', 'weights')
         return os.path.join(home, '.raymol', 'weights')
--- a/swiftui/PyMOLViewer.xcodeproj/project.pbxproj
+++ b/swiftui/PyMOLViewer.xcodeproj/project.pbxproj
@@ -12,6 +12,7 @@
 		047FD10B437F95938AB4B3B0 /* WeightsFetchStateTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0C4E3216DAFC42C515FD0C3E /* WeightsFetchStateTests.swift */; };
 		057F3C59DACA26AAC603702E /* AppBuildInfoTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6AE160C6EB58F631604002E6 /* AppBuildInfoTests.swift */; };
 		0C0E67D9436CC6DF29DE983C /* DesignEditingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 05E6ECE945556A6D1FAE98F9 /* DesignEditingTests.swift */; };
+		0C7AFC549798A22C0FE22112 /* PredictCompactPanel.swift in Sources */ = {isa = PBXBuildFile; fileRef = C2C2F36B9B6E660DD0256EC2 /* PredictCompactPanel.swift */; };
 		0DDA865CB0C8ACAC1BFD0FE9 /* NotesInspectorView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 35DC456EDC3E55F84506C8B7 /* NotesInspectorView.swift */; };
 		0F6C6BD0DB9B077E51D1EFAF /* WhatsNewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3F5377EE69A856479317FD61 /* WhatsNewModel.swift */; };
 		0FFCBEB460CD541E932326AB /* Theme.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7D3E805AC2D63B01AB7EE29C /* Theme.swift */; };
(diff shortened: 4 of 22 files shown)


javierbq:
## On-device session 1 — what ran, what broke, what is still unmeasured

Ran the ported tool on the target iPhone 15 Pro (iOS 26.6). The port works on hardware.
**Timing and peak footprint are still not measured** — the device was locked for the rest
of the session, and I am not raising `iOSMaximumTokens` on anything but a real run.

### Verified on device

Dumped from the running app rather than inferred:

```json
{"host": "1", "runtimes": ["boltz"], "platform": "ios",
 "weights_root": ".../Library/Application Support/RayMol/weights",
 "residues": 110}
```

- `RAYMOL_PREDICT_HOST` reaches `os.environ` — the "must be set before
  `Py_InitializeFromConfig`" ordering holds on iOS.
- iOS advertises **boltz only**. A Protenix predictor was refused with exactly the
  intended message: *"protenix-v2-int8 needs the 'protenix' inference runtime, which this
  build of RayMol does not carry (it has: boltz)"*.
- `PREDICT:submit:<id>` fired: a 110-residue job **passed `PredictSizeGuard` against the
  real `os_proc_available_memory()` budget** and was handed to `BoltzJobManager`. That is
  the central design question and the answer is yes.
- The Predict tool was opened on the phone (`PREDICT_FORM:ready`) — `PredictCompactPanel`
  renders and the form round-trips.
- **No jetsam kill and no crash** in any Release run.

### Bug found and fixed (commit 2)

`WeightCache.default_root()` branched on `sys.platform == 'darwin'`; CPython reports `ios`.
iOS fell through to `~/.raymol`, and `$HOME` there is the app data container ROOT — readable
but not writable. The 529 MB fetch died at its first `mkdir` with EPERM, *after* the user had
asked for it. Now under `Library/Application Support/RayMol/weights`. Verified on device;
unit test added.

### Two findings that are NOT fixed here

**1. Debug builds cannot launch on iOS.** `EXC_BAD_ACCESS`, "Thread stack size exceeded due
to excessive recursion", in `ContentView.exportMenuContent.getter` ← `iosToolPills()`, during
first layout — before any Predict view is instantiated. iOS gives the main thread 1 MB where
macOS gives 8 MB, and `-Onone` SwiftUI frames overflow it. **Release runs fine**, and all
device verification above is from Release builds. This is untouched code and looks
pre-existing, but I did not bisect against `master`, so I am not claiming it either way.
Worth its own issue.

**2. The weight download does not survive interruption.** It stalled at 101 MB / 529 MB when
the debug tunnel dropped and never resumed, even after the app was foregrounded again.
`PredictScreenAwake` stops iOS *auto*-locking, but a Python-thread `urllib` transfer dies on
suspension regardless. A 529 MB first-run fetch on a phone wants a background `URLSession`.
Worked around for testing by sideloading the bundle (SHA-256 verified against the manifest),
so it is not blocking, but it will bite a real user.

### Harness note for whoever picks this up

Do not poll a job from `PYMOL_AUTOCMD`. That runs on the main thread, and the Swift host
learns about a job only through `pollFeedback` scanning the feedback line for the `PREDICT:`
marker — also main-thread. A sleep loop there prevents the very dispatch it waits for: the
job submits, prints its marker, and nothing ever reads it. Poll from a worker thread. (Same
failure family as #284.)

### Still outstanding

- [ ] Wall time and peak `phys_footprint` for a ~110-residue single-chain fold.
- [ ] The `os_proc_available_memory()` reading on device, to confirm the ~2.5 GB budget the
      iOS fit was reasoned against.
- [ ] Raise `iOSMaximumTokens` to the largest size that actually completes — or leave it at 117.

javierbq:
## On-device session 2 — it folds, and the memory fit was wrong

**A 110-residue single chain, no MSA, folds on the iPhone 15 Pro in 39 s at a 1.72 GB peak,
with a real structure in the object panel and no jetsam kill.** That is the acceptance
criterion met.

Then a sweep, which changed the guard.

### Measured (iPhone 15 Pro, iOS 26.6, Release, single chain, no MSA, 3 recycles / 200 diffusion)

| residues | wall  | MLX peak | peak phys_footprint |
|---------:|------:|---------:|--------------------:|
| 110      |  39 s |  1.38 GB |          **1.72 GB** |
| 130      |  64 s |  1.75 GB |          **2.00 GB** |
| 164      | 250 s |  2.00 GB |          **2.35 GB** |
| 200      | refused by the guard — *"at most about 180 fit"* | | |

`os_proc_available_memory()` on that device with a session open solves to **~3.27 GB**,
back-derived from that refusal. Higher than the 2.5 GB the first fit assumed.

### The fit was measuring the wrong thing

The iOS constants were fitted against **MLX's high-water mark**. `MLX.Memory` excludes its
own buffer cache, so it under-reports the process by 300–350 MB at these sizes — while
`os_proc_available_memory()`, the budget it gets compared against, is denominated in
`phys_footprint`. Comparing the two silently favours the optimistic one.

Estimated 1.68 / 1.89 / 2.26 GB against measured 1.72 / 2.00 / 2.35 — **below every single
measurement.** The same optimistic failure the type doc records three times, arriving a
fourth time by a new mechanism, caught by data rather than by review.

It is only visible if the footprint is **sampled**. A single `phys_footprint` read after
inference is post-release and returned **1.05 GB for both the 110- and the 164-residue
run** — a reading that is obviously not the peak of anything, once you see it next to a
2.00 GB MLX peak. `BoltzJobManager.FootprintSampler` now polls every 200 ms and keeps the
max, and `Status` carries `footprint_bytes` beside `peak_bytes` because the gap between
them is not a constant.

Refitted to the footprint column: **850 MiB intercept, 7.5 MB/token, 21.5 kB/token²**,
clearing every measured point by **11–15%** (1.15× / 1.11× / 1.15×). Deliberately modest —
three real points across the range of interest justify far less padding than one point did,
and `warnFraction` withholds a further 25% on top.

`iOSMaximumTokens` **117 → 164**, the largest input actually folded to completion. Worth
noting 164 took 250 s against 99 s for the identical input in an earlier run — a 2.5×
spread that smells like thermal or memory-pressure throttling, and is its own argument
against treating that size as comfortable.

### A correction to session 1

I claimed reusing the macOS constants on iOS was "not merely conservative, unusable". That
overstated it, and the measured budget does not support it. The Mac curve **over-predicts
the device footprints by 1.23–1.39×** — safe, never a jetsam risk. Its real problem is
being over-*conservative*: at 130 residues it refuses (77% of budget) a fold that peaked at
2.00 GB and finished in 64 s. The iOS terms buy that headroom back without giving up margin
over any measurement. The test now pins that distinction rather than the overclaim.

### Net effect on what the tool accepts

On this device: 110 and 130 proceed (`.warn` proceeds on the Python path; the UI asks for
one confirmation), 164 is refused, and a refusal now names ~146 residues as the largest
that fits. That covers the 100–150 target at the low end and refuses the top honestly.

360 macOS unit tests pass. Both slices build.

### Still open

- No MSA fold has been measured on device — `bytesPerTokenMSARow` is deliberately NOT
  rescaled for iOS, so alignments are charged at the conservative Mac rate.
- The two non-blocking issues from session 1 stand: Debug builds cannot launch on iOS
  (1 MB main-thread stack), and the weight download does not survive interruption.

javierbq:
## On-device session 3 — the screen hold did not work

Running the shipping build caught a bug in this PR's own backgrounding fix.

A 110-residue fold reached **diffusion step 42 of 200 at 23.7 s and then stopped
advancing** — six consecutive status polls, identical. The phone locked and iOS suspended
the app mid-inference. That is precisely the failure `PredictScreenAwake` was added to
prevent, so the feature did not work.

**Cause: what it keyed off.** The banner and ContentView's `.onChange` both observe
`engine.predictionJobs`, which is populated by the **~500 ms object-panel poll** — a
UI-shaped feed that learns a job exists only after Python has written it and the next tick
has read it back. Between submitting a fold and pinning the display there is a window of up
to about a second, and a phone whose idle timer is nearly expired locks inside it. Nothing
about that window is observable in a simulator or on a Mac; it took a real device with a
real Auto-Lock setting.

`BoltzJobManager` knows a job has started at the instant it starts one, so it now takes the
hold itself in `run(_:)`, released through the same `defer` as the footprint sampler.

The view path stays — it also covers the 529 MB weight download, which has no
`BoltzJobManager` job. The two compose through a **count** rather than a boolean, so:

- neither can release the other's hold (`apply(_:)` can now only raise, never drop one a
  job owns);
- overlapping folds (`n_models > 1`, queued jobs) keep the display pinned until the last
  one finishes;
- releases are floored at zero, because an unbalanced release driving the count negative
  would make the next genuine hold a silent no-op and reintroduce this exact freeze.

Three tests cover the counting, including the negative-floor case. **363 macOS unit tests
pass; both slices build.**

### Note on the earlier measurements

The sweep numbers in session 2 are unaffected — those folds ran to completion with the
phone awake, and the guard constants are fitted to their footprints. This bug governs
whether a run *survives*, not what it costs.

### Still to confirm

The post-fix device run is queued behind the phone being unlocked; the fix is verified by
unit test and by construction, not yet by a second on-device fold. Worth re-running a
110-residue fold with the phone left untouched to confirm it now survives past step 42.

javierbq:
## Confirmed on device — the fold now completes

Re-ran the 110-residue single-chain fold on the iPhone 15 Pro with the job-owned screen
hold in place. **It cleared step 42 and ran to completion.**

```
residues   110          state      done
wall       40.8 s       elapsed    38.9 s
peak       1.41 GB      footprint  1.60 GB
object     boltz2_prediction_de6ee3cd
```

Structure pulled off the device and checked, so this is not just a status flag:

```
atoms 877 · residues 110 · CA 110 · chain A
bbox  31.4 x 28.7 x 40.8 A
pLDDT (B-factor) min 30.90 · mean 46.15 · max 59.61
```

877 atoms for 110 residues, one CA per residue, and a compact ~31x29x41 A envelope — a
folded single chain, not a collapsed or extended artifact.

Both Python fixes were verified present inside the installed bundle before the run (three
`check_available` call sites in `appkit_predict.py`, the iOS refusal in `boltz2_bf16.py`),
so this build is the one carrying them.

### On the numbers

This run peaked at 1.41 GB (1.60 GB footprint) against the 1.72 GB the guard was fitted
to for 110 residues. Lower than the fit, so the guard stays conservative and needs no
change — but it is a reminder that peak varies run to run and the fit should track the
high end of a sweep, not a single sample.

### Caveat worth stating plainly

**Mean pLDDT 46 is low confidence.** That is expected for a single sequence with no MSA
and is not a bug in the port — the target for this work was explicitly "shallow or no
MSA". It does mean nobody should read this particular model as a useful structure. What is
demonstrated is that the pipeline runs end to end on a phone and returns a well-formed
result with per-atom confidence; prediction QUALITY at this MSA depth is a separate
question, and no MSA-backed fold has yet been measured on device.

Issue #342, opened 2026-08-23
De novo binder design via rfd3-mlx: the first method that generates a chain rather than folding one


javierbq:
## Summary

Add **de novo binder design** as an on-device capability, via the MLX-Swift port of
RFdiffusion3 in [`javierbq/rfd3-mlx`](https://github.com/javierbq/rfd3-mlx) (`RFD3Kit`).

Scope is deliberately one mode: given a target structure and a set of hotspot residues,
generate a new binder backbone against it. Sequence redesign, refolding, interface gating
and any run-until-N-pass loop are **out of scope here** — see *Scope* below.

The port is in better shape than protenix-mlx was at #309: the network is complete, the
featurizer is already in Swift, it designs end to end against a real target, and no MLX
type leaks through the public surface. The work is integration, not porting.

## What already exists

`javierbq/rfd3-mlx` — a `.library("RFD3Kit")` SwiftPM product, macOS 14 / iOS 17, mlx-swift
pinned `exact: 0.31.6` (the same pin MPNNKit and boltz-mlx already hold).

| Layer | State |
| :--- | :--- |
| Swift engine (token encoder/init, diffusion transformer, decoder, sampler, sequence head) | working |
| Swift featurizer (`Featurizer.binderDesign`, motif/hotspot conditioning) | working, **in Swift** |
| End-to-end design (target PDB → binder backbone + sequence) | working |
| Measured peak-memory model + pre-GPU refusal (`RFD3Budget`, `preflight`) | working |
| Weight pack format + sha256 manifest + provenance (`RFD3Pack`) | working |
| Published tag / release pack | **neither exists** — see *Package hygiene* |
| Motif scaffolding, symmetry, small-molecule binders, partial diffusion | not exposed |

The whole surface RayMol would consume:

```swift
public struct RFD3Model {
  public init(packDirectory: URL, verifyChecksums: Bool = true) throws
  public static func autoTarget(fromPDB: String) -> (residues: [Residue], hotspots: [Int])?
  public static func preflight(target: [Residue], options: Options = Options(),
                               budgetBytes: Int? = nil) throws -> FeatSet
  public func designBinder(target: [Residue], options: Options) throws -> Result
}
// Options: binderLength, hotspots, numTimesteps, nRecycle, seed, memoryBudgetBytes,
//          onProgress: ((Int, Int) -> Void)?
// Result:  pdb (chain A = target, held fixed; chain B = designed binder),
//          binderSequence, stats, seconds
[...]

Issue #291, opened 2026-08-14
Running predictions need an ambient progress indicator: today a job is invisible unless the object panel is open, and always invisible on iOS


javierbq:
**Is your feature request related to a problem? Please describe.**

A structure prediction runs for **seconds to tens of minutes** (a single 600-residue run is
~11 minutes), it is asynchronous by design, and `n_models` fans out into up to 20 separate
jobs. During all of that, the app gives the user essentially nothing to look at.

The complete set of running-job feedback in RayMol today is:

1. a placeholder row in the object panel whose enable-toggle is greyed out
   (`ObjectPanel.swift:2058`), and
2. `.help(entry.pendingDetail ?? "")` — a **hover tooltip** (`ObjectPanel.swift:2196`).

Which means a running prediction is invisible whenever any of these is true:

- **The object panel isn't open.** Nothing else in the app mentions the job.
- **You're on iOS/iPadOS.** `.help` is a no-op there, so there is no phase, no percentage,
  and no cancel — just an empty object that eventually fills with atoms.
- **You aren't hovering the exact row.** Even on macOS with the panel open, progress is
  hidden behind a mouse-over, so "is it still going, or did it die?" costs a hover every
  time you want to know.

With more than one job outstanding there is no aggregate anywhere: no "3 of 5 models done",
no combined progress, no single place that answers "is RayMol busy predicting right now?"

**The data is already flowing app-wide — there is simply no view for it.**
`PyMOLEngine.pollObjects()` (`PyMOLEngine.swift:3050`) runs `_ai.poll_panel()` every 5th
tick of the 100 ms timer (~500 ms, ~2 Hz) and is **not** gated on panel visibility, so the
`pending` map from `appkit_inspector._pending_map()` (`appkit_inspector.py:440`, emitted at
`:505`) arrives continuously whether or not the object panel is on screen. Its only consumer
is `ObjectPanel`. This is a presentation gap, not a plumbing gap.

**Describe the solution you'd like**

An **ambient, non-blocking progress indicator** that appears whenever ≥1 prediction job is
running, visible independently of which panel is open — on **both** macOS and iPadOS.

`MCPDrivingBanner` is the right shape and the right precedent: a slim top banner with a
status dot, a live one-line detail, and a Stop button, that does not block interaction. It is
currently `#if os(macOS)` only, so this needs an iOS treatment either way.

Content, all of which is already available:

- **Aggregate state across jobs.** `predicting.pending_objects()` returns
[...]

PR #344, opened 2026-08-23
fix(predict): write a Protenix job's terminal status before discarding its placeholder

javierbq:
`predicting.discard_pending` re-reads the status file **fresh** to decide whether to retain a
failure card in `_RECENT` — it can, because nothing ever deletes that file. So the order is
load-bearing: a discard that lands first pops `_PENDING`, the only map that could observe the
outcome, and the failure is stranded. The card silently vanishes and a run that produced
nothing explains nothing.

`BoltzJobManager.settle` exists to encode that order and has had a test since #291.
`ProtenixJobManager` never used it — all four of its terminal paths did the two steps inline,
discard first.

## Three of the four are a real race, and it isn't the one that looks worst

`discardPlaceholder` hops to the **main** queue while `writeStatus` is synchronous on the
calling thread. That makes the threading decisive:

| site | thread | effect today |
|---|---|---|
| `submit` refusal | main | hop is deferred past the write — accidentally harmless |
| `run` cancel-during-featurize | background | main thread can take the discard mid-write |
| `run` `catch .cancelled` | background | same |
| `run` `catch` (failure) | background | same |

So: a latent hazard at the one site that reads worst, and a live race at the three that
carry a cancellation or a failure *after minutes of folding* — precisely when a card saying
why is worth having.

The fix is the same at all four: route them through the shared helper instead of repeating
the sequence. A second copy of an ordering is a second chance to get it backwards, which is
what happened here. `run` gains a local `settle(_ state:_ phase:error:)` mirroring
`BoltzJobManager.run`'s own, so the two managers now settle identically.
`BoltzJobManager.settle` drops `private` and gains no new behaviour.

## Tests

Both were confirmed to **fail against the unfixed code** before the fix went in — 6 failures
naming all four call sites:

- `testARefusalWritesStatusBeforeDiscardingThePlaceholder` — drives `submit` with a
  100k-residue request (refused before MLX or any weight read) and asserts the `settleTap`
  sequence is exactly `["write", "discard"]`. Against the old code the tap recorded `[]`:
  the manager never entered `settle` at all, which is the finding stated as an assertion.
- `testNoTerminalPathDiscardsThePlaceholderItself` — greps the manager's own source for
  zero direct `discardPlaceholder` calls, plus at least one `settle` so it can't be
  satisfied by deleting the teardown. A grep because the three paths inside `run` need a
  real 214 MB pack and a real fold to reach; same discipline as
  `predict_api.testEveryMessageHelperUsedByPredictingExists`.

**431 Swift tests, 0 failures.** Both slices compiled (macOS + iOS) — no CI job compiles
Swift.

The regenerated `project.pbxproj` is deliberately **not** included: no files were added, so
its only diff was fresh random UUIDs.

## Note for whoever merges second

This widens `BoltzJobManager.settle` from `private static` to `static`. #343 edits the same
declaration line (making it `static` **and** adding `pythonModule: String = "predicting"`),
so whichever lands second gets a one-line textual conflict. The resolution is always #343's
signature — it is a superset. Everything else here is in `ProtenixJobManager.swift` and
`ProtenixRuntimeTests.swift`, which #343 does not touch.

Noticed while implementing #342 and kept out of #343 so that change did not touch a shipped
predictor's behaviour.

### DIFF
--- a/swiftui/PyMOLViewer/Shared/BoltzJobManager.swift
+++ b/swiftui/PyMOLViewer/Shared/BoltzJobManager.swift
@@ -285,7 +285,11 @@ final class BoltzJobManager {
     /// run that failed used to just make its object vanish.
     ///
     /// One function rather than six call-pairs so a seventh exit cannot get it wrong.
-    private static func settle(_ request: Request, _ status: Status, to url: URL) {
+    /// Internal rather than private: `ProtenixJobManager` settles its own jobs and must use
+    /// THIS function rather than a second copy of the ordering. That is the whole point --
+    /// a second copy is a second chance to get it backwards, which is exactly what had
+    /// happened at all four of that manager's terminal paths.
+    static func settle(_ request: Request, _ status: Status, to url: URL) {
         try? writeStatus(status, to: url)
         #if DEBUG
         settleTap?("write")
--- a/swiftui/PyMOLViewer/Shared/ProtenixJobManager.swift
+++ b/swiftui/PyMOLViewer/Shared/ProtenixJobManager.swift
@@ -52,9 +52,8 @@ final class ProtenixJobManager {
     /// Refuse or accept a submitted request. Runs on the MAIN thread.
     func submit(_ request: BoltzJobManager.Request) {
         if let failure = Self.preflight(request) {
-            BoltzJobManager.discardPlaceholder(request)
-            try? BoltzJobManager.writeStatus(
-                failure, to: URL(fileURLWithPath: request.statusPath))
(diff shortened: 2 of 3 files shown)


javierbq:
Structural follow-up filed as #345: the three inference managers are in a dependency cycle, and the wire types plus the whole job shell live inside `BoltzJobManager`. Deliberately queued behind this PR rather than folded into it — doing it now would put three PRs in flight over one file. This PR is unaffected.

javierbq:
BAD

PR #374, opened 2026-09-03
feat(cartoon): spline tessellation mode + artifact-free traced shadows

javierbq:
## Summary

Spline cartoon tessellation, artifact-free real-time ray-traced shadows/AO, and new defaults. Six commits; each is self-contained.

### 1. `cartoon_spline` — ChimeraX-style path tessellation (now the default)

- One natural cubic spline per cartoon segment through the (flattened) control points, `sampling` points per interval on the spline; orientation frame carried by parallel transport with a sigmoid twist so it lands on the next residue's orientation vector.
- Replaces only `CartoonGenerateSample` + `CartoonGenerateRefine`; extrusion, arrows, flat sheets, putty, cylindrical helices and nucleic acids are shared. Cyclic segments fall back to the classic sampler. In spline mode the strand/arrow strips get true saddle normals (`ExtrudeTwistNormals`) and AUTO sampling is 20 rings/residue for < 100k atoms.
- `set cartoon_spline, off` restores the classic tessellation exactly. Exposed as **Spline ribbon** in the object panel and inspector; `set` invalidates the rep.
- `examples/devel/spline_cartoon.py` is the pure-Python prototype the C++ port was checked against.

### 2. Traced shadows / AO without acne, stripes or sawtooth

Root cause of every artifact: the post passes only have a depth-reconstructed normal, and on a twisted ribbon quad the two triangles tilt alternately, so anything oriented from that normal flickers per triangle. `rt_ao` now:

- finds the triangle each pixel lies on with a primary ray and uses its strip-smoothed facet normal for the AO hemisphere/origin, the shadow-ray origin and the self-hit tests (sphere impostors keep the analytic depth normal);
- shadow ray: own-face rule (parallel facet within the slab reach), near-range terminator skip (< 3 Å), else caster; facing gate only for nearby sphere casters via a flag channel (`_rtAO` is RGBA16Float);
- AO: any-hit, 0.3 Å base skip, longer skip for grazing rays;
- world-tri vertex buffer bound to the shader (`_rtTriBuffer`, `RTU.triInstance/triCount`); sphere proxy 320 tris, 12-side cylinders.

### 3. Performance (`RAYMOL_GPU_TIMING=<file>` diagnostic added)

| 4 Mpx window, rotating 2kpp | GPU ms / frame |
|---|---|
| RT off | ~7–8 |
| RT + traced shadows, full res | ~28–31 |
| `metal_rt_scale 0.5` (default) | ~16 |

New `metal_rt_scale` (Display panel → RT resolution) renders the RT pass at a fraction of the window; exports stay full-res. A static scene no longer re-renders on every UI poll (`PyMOLBridge_RunPythonQuiet` for the five periodic polls): 0 frames in 10 s idle, was ~25.

### 4. iOS launch crash under Metal validation

MSAA depth-stencil resolve is unsupported on the simulator / pre-A12 GPUs → assert in `beginFrame` on the first frame. `setSampleCount` falls back to single-sample there. (Also merged separately as #375.) RT/Post uniform structs padded to 16-byte multiples.

### 5. New defaults

`cartoon_spline` on, `metal_raytrace` on (runtime-gated by RT support), `metal_rt_shadows` on, `metal_temporal_ao` on, `metal_rt_samples` 32; `metal_rt_scale` 0.5, AO 5 Å / 0.72, shadow 0.45, MSAA on unchanged. All built-in themes carry `rayTrace: true, shadows: true` (the theme re-asserts these on launch); custom themes decode to the same; `raymol_theme.set_palette` defaults `ray_trace=True`. iOS note: same defaults on RT-capable devices — gate `rayTrace` per platform in `Theme.swift` if too heavy.

## Verification

- `testing/tests/settings/cartoon_spline.py` (6 tests) + settings / side-chain-helper / theme_apply / theme_preview / appkit_settings suites pass against a CLI build of this tree.
- Exercised in the app on 2kpp, 1ehz (RNA), 1oky (gaps, skips, all helix modes, putty), a synthetic cyclic peptide, 1aon (59k atoms).
- Metal RT verified by screenshotting the live window across an 8-orientation grazing-light sweep, close-ups with spheres/sticks, and with `cartoon_flat_sheets` off: no per-triangle acne, no striped bands, no comb on shadow edges.
- Fresh launch reads back all new defaults; `fetch 2kpp` renders spline + RT without any `set`.
- iOS simulator (iPhone 17 Pro / iPad Pro M4) launches under `MTL_DEBUG_LAYER=1` with RT + spline on.


javierbq:
**Performance follow-up (GPU load with RT shadows on).**

Added a `RAYMOL_GPU_TIMING=<file>` diagnostic (frames/s + GPU ms/frame) and measured on a 1332×771 pt window rotating 2kpp:

| configuration | GPU ms / frame |
|---|---|
| RT off (SSAO + shadow map) | ~7–8 |
| RT on, AO 16 spp + traced shadows, full res | ~28–31 |
| same, `metal_rt_scale 0.5` (new default) | ~16 |
| same, `metal_rt_scale 0.25` | ~13 |

The cost is the full-resolution AO pass itself; traced shadows add ~6 ms on top. New `metal_rt_scale` (Display panel → "RT resolution") renders the RT AO/shadow pass at a fraction of the window; the composite's bilinear upsample + depth-aware blur hides it, and exports stay full-res. The AO query is back to any-hit and the sphere proxy is 320 tris.

Also in this branch since the last update: no ring stripes on twisted strands with `cartoon_flat_sheets` off (primary-ray facet normal for AO/shadow self-tests, AO terminator guard, true saddle normals on strand quads in spline mode).

Not changed: a static scene still re-renders ~2–3 fps because the object-panel poll goes through the command path and sets the redisplay flag — with RT on that is ~5–10% GPU while idle. Worth a separate fix.

javierbq:
**Idle re-render fixed too.** The periodic UI polls (object panel, detail card, movie, sequence) run read-only Python that creates temporary selections; the core marks the scene changed for those, so a still window re-rendered 2–3 full frames/s. `PyMOLBridge_RunPythonQuiet` restores the redisplay flag after a poll if it was clear before; the five polls use it. Static view with RT + traced shadows: 0 frames in 10 s (was ~25). Interaction and real commands still render, and the panel keeps updating.

javierbq:
**New defaults (per Javier):** `cartoon_spline` on, `metal_raytrace` on (still gated by device RT support), `metal_rt_shadows` on, `metal_temporal_ao` on, `metal_rt_samples` 32; `metal_rt_scale` 0.5, AO radius 5 Å / 0.72, shadow 0.45, MSAA on unchanged. All four built-in themes now carry `rayTrace: true, shadows: true` (the theme re-asserts these on launch), custom themes decode to the same, and `raymol_theme.set_palette` defaults `ray_trace=True`. Fresh launch reads back all of the above. Note for iOS: the same defaults apply on RT-capable iPhones/iPads; if that is too heavy there, gate `rayTrace` per platform in `Theme.swift`.

javierbq:
**iOS simulator test (iPhone 17 Pro + iPad Pro M4, Metal API validation asserting).**

- Fresh launch reads back the new defaults: `cartoon_spline` on, `metal_raytrace` on, `metal_rt_shadows` on, `metal_temporal_ao` on, 32 spp, `metal_rt_scale` 0.5. MSAA falls back to single-sample on the simulator (the #375 path).
- Cartoon-only: launches, renders, RT export OK, no assertions, on both devices.
- Sticks exposed a **pre-existing** validation abort in `drawCylinderImpostors`: the CylU / SphereU C++ mirrors were 8 bytes short of the MSL struct size (200 vs 208, 184 vs 192). Fixed in 9fdfc24 (explicit tail padding both sides; every other bound uniform struct audited and already matching).
- After the fix: cartoon + side-chain sticks + spheres with RT AO and traced shadows renders and exports on iPhone and iPad with zero validation assertions and no crash reports.

PR #409, opened 2026-09-06
perf(pick): do the hover/click atom pick in C++ (#394)

### LINKED ISSUE #394: Hover pick rebuilds a chempy model of every drawn atom on every pick (~1.5 us/atom at up to 22 picks/s): main thread saturates on large structures
## Summary

Every hover pick rebuilds a `chempy` model of **all drawn atoms of every enabled object** (`cmd.get_model`) and projects them in pure Python. Hover preview and hover readout are both on by default, and the pick is throttled to ~22/s, so on a large structure the main thread is saturated while the pointer moves over the viewport.

## Evidence

- [`metal_pick._pick_atom`](https://github.com/javierbq/RayMol/blob/62167b29e/modules/pymol/metal_pick.py#L395-L409): per enabled object, `sel = '(%s) and (%s)' % (obj, _DRAWN_REPS)` then `model = cmd.get_model(sel)` (twice per object when the displayed state is >1), followed by the per-atom Python projection loop.
- Called from [`hover_preview_at`](https://github.com/javierbq/RayMol/blob/62167b29e/modules/pymol/metal_pick.py#L700) ← [`PyMOLEngine.hoverPreview`](https://github.com/javierbq/RayMol/blob/62167b29e/swiftui/PyMOLViewer/Shared/PyMOLEngine.swift#L3183) via `runPython`, leading-edge throttled at [`kHoverInterval = 0.045`](https://github.com/javierbq/RayMol/blob/62167b29e/swiftui/PyMOLViewer/Shared/PyMOLEngine.swift#L3168). Both [`hoverPreviewEnabled` and `hoverReadoutEnabled`](https://github.com/javierbq/RayMol/blob/62167b29e/swiftui/PyMOLViewer/Shared/PyMOLEngine.swift#L287-L300) default to `true`.
- No coordinate cache exists anywhere in `metal_pick.py`. `pick_at`, `hover_design_at` and `appkit_measure` share the same path.

**Measured** in the live app:

| drawn atoms | `_pick_atom` | `hover_preview_at(…,1,1)` |
|---|---|---|
| 1,803 | 1.7 ms | 2.1 ms |
| 19,833 | 29.5 ms | 31.0 ms |

≈1.5 µs per drawn atom (`cmd.get_model('2kpp')` alone is 4.1 ms for 1,803 atoms). At 50k drawn atoms that is ~75 ms per pick at up to 22 picks/s: more than 100% of the main thread, with the 120 Hz render loop, the 10 Hz feedback timer and the panel polls queued behind it. Each pick also allocates and frees ~N Python `Atom` objects.

## Suggested fix

Do the pick in C++: a `_cmd` entry that walks each `CoordSet`'s coordinate array (honouring `visRep` bits, TTT and grid slot) with the current view and returns the best atom index, or reuse the renderer's depth/ID buffer. A Python-side coordinate cache is blocked by the absence of numpy in the bundle. Keep `metal_pick.py` as the thin policy layer (selection-mode expansion, `_preselect`, readout payload).

Effort: M. Risk: medium (grid-mode and moved-object handling must be ported exactly; `test_metal_pick.py` covers them).

---
_From the 2026-09-05 tech-debt & performance review (`docs/reviews/2026-09-05-tech-debt-and-performance.md`, to be added in a follow-up PR). Measurements taken on the `RayMol-PR370` dev build (2026-09-01) on Apple silicon; treat them as order-of-magnitude._

javierbq:
Fixes #394.

## What was slow

Every pick rebuilt a `chempy` model of **all drawn atoms of every enabled object** (`cmd.get_model`) and projected them in pure Python. Hover preview and hover readout are both on by default and the pick is throttled to ~22/s, so on a large structure the main thread had nothing left for the render loop, the feedback timer or the panel polls.

## What changed

The per-atom work moves to `layer3/MetalPick.cpp`, reached through a new `_cmd.metal_pick`. It:

- walks each `CoordSet`'s coordinate array directly (`IdxToAtm` → coord), so there is no chempy model and no per-atom Python object;
- tests the drawn reps as a `visRep` bitmask instead of running the selection parser once per object — including the cartoon/ribbon special case, which is `(cartoon|ribbon bit) && (flags & cAtomFlag_guide)`;
- applies the same state matrix + object TTT that `CoordSetGetAtomTxfVertex` does, so a moved object still picks where it renders;
- honours the clip slab and the front-most-of-a-cluster rule exactly as before.

`modules/pymol/metal_pick.py` stays the policy layer: camera parsing, grid-cell mapping, `mouse_selection_mode` expansion, `_preselect`, readout payload. The Python projection is kept as `_python_pick` — it is the reference the native path is tested against, it is what the `PYMOL_PICKDEBUG` harness runs (it reports the candidate count and projected extent the C++ path doesn't collect), and it is the fallback for cores built without the new entry point (e.g. the shadow-tree setup over Homebrew pymol).

`pick_at`, `pick_info_at`, `hover_preview_at`, `hover_design_at`, `metal_move` and `appkit_measure` all go through `_pick_atom` and pick this up unchanged. No Swift change.

## Measured

50,400 drawn atoms (84 copies of a 60-mer, sticks + spheres), 800x600 viewport, Release core on Apple silicon:

| | before | after |
|---|---|---|
| `_pick_atom` | 450–920 ms | **0.32 ms** |
| `hover_preview_at(…, 1, 1)` | ~same + ~5 ms | **6.0 ms** |

At 18,000 drawn atoms it is 72 ms → 0.10 ms.

The remaining hover cost is `cmd.select('_preselect', …)` at 5.3 ms (the readout payload is 0.1 ms) — a selector evaluation over the whole atom table, unrelated to this path and left for a follow-up. At 22 picks/s that is ~15% of the main thread rather than the >100% the issue measured.

## Behaviour change worth flagging

An object with **fewer states than the current frame** is no longer pickable. The renderer draws nothing for it (`StateIterator` clamps to `[0, nstate)`), while the old code fell back to `get_model()`'s default state 1 and picked it anyway — i.e. it could select an object that wasn't on screen. The native path asks each object for its own current state (`CObject::getObjectState`), which is also what the renderer asks.

## Tests

New `testing/tests/raymol/metal_pick.py` (22 cases, live core, camera pinned with `set_view` the same way `api/box_select.py` does it):

- every drawn rep is pickable and agrees with the `_DRAWN_REPS` selection; `labels` alone is not — this is the guard against the bitmask and the selection string drifting apart;
- a cartoon-only object resolves every hit to a guide atom, never to an invisible side chain;
- clip slab, front-most-of-two-overlapping, hidden reps, disabled objects, moved (TTT) objects, displayed state, the wider object-pick radius;
- `TestNativeParity` runs the C++ pick and the Python reference over an 11x11 grid of cursor positions on a mixed scene and requires the same hit.

Verified the suite fails without the fix: dropping a rep from the mask fails its `testEachDrawnRepIsPickable` case, and folding cartoon/ribbon into the plain drawn mask fails the guide-atom test and both parity tests.

`testing/tests/test_metal_pick.py` (headless, mocked `cmd`) now covers both branches — the existing case pins the Python path, a new one pins the native path and asserts it never touches `get_model`.

Added to the embedded-tests workflow. Full CI list green locally: 955 unittest + 287 pytest, plus `api/box_select.py`, `api/selecting.py`, `api/querying.py`.

Both build configurations compile clean (no new warnings): the `pip install .` GLUT/OpenGL extension and `swiftui/build_macos.sh`'s Metal-only `libpymol_core.a`.

### DIFF
--- a/.github/workflows/raymol-embedded-tests.yml
+++ b/.github/workflows/raymol-embedded-tests.yml
@@ -74,6 +74,7 @@ jobs:
               testing/tests/raymol/design_enumerate.py \
               testing/tests/raymol/design_region.py \
               testing/tests/raymol/design_saverestore.py \
+              testing/tests/raymol/metal_pick.py \
               testing/tests/raymol/scene_ttt.py \
               testing/tests/raymol/shadow_extent.py \
               testing/tests/predict \
--- a/layer3/MetalPick.cpp
+++ b/layer3/MetalPick.cpp
@@ -1,0 +1,158 @@
+/*
+ * Screen-space atom picking for the Metal backend -- see MetalPick.h.
+ */
+
+#include "MetalPick.h"
+
+#include <algorithm>
+#include <cfloat>
+
+#include "AtomInfo.h"
+#include "CoordSet.h"
+#include "ObjectMolecule.h"
+#include "Rep.h"
+#include "Setting.h"
+#include "Vector.h"
+
+namespace
+{
+
+/// An atom that projected inside the pick radius.
+struct PickCand {
+  float d2;
+  float depth;
+  ObjectMolecule* obj;
+  int atm;
+  float sx, sy;
+};
+
+/**
+ * True if `ai` has geometry a pick can land on.
+ *
+ * Cartoon and ribbon are special: showing them ORs their bit onto EVERY atom of
+ * the object (side chains and solvent included), but the spline only runs
+ * through the guide atoms, so only those are hittable. Mirrors the _DRAWN_REPS
+ * selection in pymol/metal_pick.py.
+ */
+inline bool is_drawn(
+    const AtomInfoType* ai, int rep_mask, int guide_rep_mask)
+{
+  if (ai->visRep & rep_mask)
+    return true;
+  return (ai->visRep & guide_rep_mask) && (ai->flags & cAtomFlag_guide);
+}
+
+} // namespace
+
+MetalPickHit MetalPickAtom(PyMOLGlobals* G,
+    const std::vector<ObjectMolecule*>& objects, int state,
+    const MetalPickCamera& cam, float ndc_x, float ndc_y, float max_ndc2,
+    float cluster_ndc2, int rep_mask, int guide_rep_mask)
+{
+  MetalPickHit hit;
+
+  const float r00 = cam.rot[0], r01 = cam.rot[1], r02 = cam.rot[2];
+  const float r10 = cam.rot[3], r11 = cam.rot[4], r12 = cam.rot[5];
+  const float r20 = cam.rot[6], r21 = cam.rot[7], r22 = cam.rot[8];
+  const float tx = cam.pos[0], ty = cam.pos[1], tz = cam.pos[2];
+  const float ox = cam.origin[0], oy = cam.origin[1], oz = cam.origin[2];
+  // A degenerate slab (front >= back) means the view layout didn't carry usable
+  // clip planes; culling against it would make nothing pickable at all.
+  const bool clipped = cam.clip_back > cam.clip_front;
+
+  if (!(cam.tan_half > 0.f) || !(cam.aspect > 0.f))
+    return hit;
+
+  std::vector<PickCand> cands;
+  float d2min = FLT_MAX;
+
+  for (auto* obj : objects) {
+    if (!obj)
+      continue;
+    const CoordSet* cs = obj->getCoordSet(state);
+    if (!cs)
+      continue;
+
+    // Same transform CoordSetGetAtomTxfVertex applies, hoisted out of the loop:
+    // the pick has to agree with where the renderer put the atom, and for a
+    // moved object (non-identity TTT) that is nowhere near the raw coordinate.
+    const double* mat = nullptr;
+    if (!cs->Matrix.empty() && SettingGet<int>(*cs, cSetting_matrix_mode) > 0)
+      mat = cs->Matrix.data();
+    const bool use_ttt = obj->TTTFlag != 0;
+
+    const AtomInfoType* atom_info = obj->AtomInfo.data();
+    const int n_index = cs->getNIndex();
+
+    for (int idx = 0; idx < n_index; ++idx) {
+      const int atm = cs->IdxToAtm[idx];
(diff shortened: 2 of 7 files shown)


javierbq:
## Verified on the host in a real app build

`RayMol-394.app` (Debug, `swiftui/build_macos.sh` core + `xcodebuild PyMOLViewer_macOS`), launched as its own instance on MCP port 48994.

### Correctness — 1UBQ, cartoon + sticks

| check | result |
|---|---|
| `_cmd.metal_pick` present / module using it | yes / yes |
| CA picks at residues 5, 20, 45, 60 | each resolved to its own residue |
| side-chain pick (PHE45 CG) | PHE45 CG |
| `_preselect` after hover | exactly the 11 atoms of resi 45 |
| `sele` during hover | 0 — untouched |
| hover readout JSON | `{hit, obj 1ubq, chain A, resi 45, resn PHE, name CG, mode 1, state 1, nstates 1}` |
| click | commits resi 45 into `sele` |
| empty space | preview cleared, readout `{"hit": false}` |
| `pick_info_at` (iOS long-press) | PHE45, `sel` correct |
| `hover_design_at` | PHE45 |
| `appkit_measure`, two atom picks | emitted 14.41 Å; `cmd.get_distance` on those two atoms is 14.41 |
| `metal_move` object radius | hits `1ubq` from 0.25 NDC off the atom |

The renderer agrees with the pick: with a synthetic cursor on PHE45's CZ, the cyan `_preselect` markers render on exactly that side chain in the Metal view.

### Cost — same binary, A/B by flipping `metal_pick._native_pick_cache`

49,500 drawn atoms (75 copies of 1UBQ, sticks + spheres), 1966x720 viewport:

| | per hover | main thread at the app's 22 picks/s |
|---|---|---|
| pre-#394 path | **358.4 ms** | **788 %** |
| this PR | **2.06 ms** (`_pick_atom` alone 0.274 ms) | **4.5 %** |

At 18,600 drawn atoms: 81.0 ms → 0.102 ms.

The saturation in the issue title reproduces exactly: while the Python sweep ran, the app stopped answering MCP altogether — two calls in a row died on the 30 s main-thread guard — and only recovered once the sweep finished. The native sweep never blocked it.

PR #431, opened 2026-09-08
fix(seq-viewer): live-highlight residues during drag-selection (#429)

### LINKED ISSUE #429: macOS: sequence viewer drag-selection doesn't highlight live — background only updates after mouse release
### Summary
In the **sequence viewer**, dragging to select a span of residues does not highlight the background live. The selection-color background appears only **after the mouse is released**. It should update dynamically during the drag to show what is being selected.

### Steps to reproduce
1. Load a structure and open the sequence viewer (Seq).
2. Press on a residue and drag across a span.
3. Observe: while dragging, the swept residues are not highlighted; the background fills in only once the mouse button is released.

### Expected
As the drag moves, the residues currently under the sweep highlight immediately (live), the same way a text selection tracks the cursor, so you can see the span you're about to select before releasing.

### Actual
No visual change during the drag; the highlight snaps in on release.

### Likely cause
The per-cell highlight is bound solely to the *confirmed* PyMOL selection, which is fetched asynchronously:
- Highlight test: `swiftui/PyMOLViewer/Panels/SequencePanel.swift:212` (`selected = engine.selectedResidueKeys.contains(r.selKey)`) → `:221` (`.background(selected ? selectionBG : .clear)`).
- Drag `onChanged` dispatches a PyMOL `select` and then schedules an async round-trip to read it back: `:303-313` → `applyToggle` → `:347-349` `DispatchQueue.main.asyncAfter(0.05) { fetchSequenceSelection() }`.

So the highlight only reflects `selectedResidueKeys` after the `select` command runs, writes the SEQSEL feedback, and is parsed back into the published set. During a fast drag that round-trip lags (and the main thread is busy handling the gesture), so nothing visibly changes until the drag ends and the pipeline catches up.

### Suggested fix
Drive an **optimistic local highlight** during the drag: keep the current swept range in `@State` (it's already computed as `lastRange` in `onChanged`), union it into the `selected` test so those cells highlight immediately, and clear/reconcile it against `selectedResidueKeys` in `onEnded`. That decouples the on-screen feedback from the async selection round-trip.

### Pointers
- `swiftui/PyMOLViewer/Panels/SequencePanel.swift:212` and `:221` (highlight binding)
- `swiftui/PyMOLViewer/Panels/SequencePanel.swift:301-313` (drag `onChanged`) and `:347-349` (async fetch)

Environment: macOS, local `origin/master` Debug build (RayMol 1.10.0).

javierbq:
Fixes #429

## Problem
In the macOS sequence viewer, dragging to select a span of residues did not highlight the background live — the selection color filled in only **after** the mouse was released.

## Root cause (confirmed)
The per-cell highlight in `SequencePanel.swift` was bound solely to the **confirmed** PyMOL selection (`engine.selectedResidueKeys`), which is only populated by an async round-trip: the drag `onChanged` dispatches a `select` command via `applyToggle`, which schedules `fetchSequenceSelection()` 0.05s later. During a fast drag that round-trip lags (and the main thread is busy with the gesture), so nothing visibly changed until the drag ended and the pipeline caught up.

## Fix
Add an **optimistic local highlight** during the drag:
- New `@State dragSelKeys: Set<String>` holds the selKeys currently under the sweep.
- `onChanged` recomputes it from the swept `range` on every update, so it tracks the cursor.
- The per-cell `selected` test unions `dragSelKeys` with `selectedResidueKeys`, so swept cells light up immediately.
- `onEnded` pulls the confirmed selection (`fetchSequenceSelection()`) and clears the overlay after a short defer — this avoids a one-frame flicker while the confirmed set catches up, and the deferred clear (guarded so a new drag can't be wiped) guarantees the overlay can never get stuck on.

The real `'sele'` behaviour (additive toggle) is unchanged. Plain click and shift-extend go through `handleClick`/`applyToggle` and are untouched; the drag gesture keeps its `minimumDistance: 8` so a plain click does not trigger it.

## Verified
- **Builds green**: rebuilt `libpymol_core.a` from this branch's source, then `xcodebuild ... PyMOLViewer_macOS -configuration Debug` → **BUILD SUCCEEDED** (the Swift compiles and the app links).
- Reviewed the click/shift/drag code paths to confirm the additive-toggle selection semantics are unchanged and the overlay clears on drag end.

## Not verified
- No interactive/on-screen run of the drag was performed in this session, so the live-highlight behaviour was validated by code inspection + a clean build, not by observing the actual sweep. There is no Swift unit test covering the `SequencePanel` drag overlay (the `test_appkit_sequence*.py` tests exercise the C++ SEQPANEL/SEQSEL backend, which this Swift-only change does not touch).

### DIFF
--- a/swiftui/PyMOLViewer/Panels/SequencePanel.swift
+++ b/swiftui/PyMOLViewer/Panels/SequencePanel.swift
@@ -93,6 +93,12 @@ struct SequencePanel: View {
     @State private var residueFrames: [String: CGRect] = [:]
     @State private var lastRange: ClosedRange<Int>? = nil
     @State private var anchorIndex: Int? = nil  // for Shift-click range
+    // Optimistic live-highlight during a drag: the selKeys currently under the
+    // sweep, unioned into the per-cell highlight test so they light up
+    // immediately, without waiting for the async PyMOL selection round-trip
+    // (issue #429). Cleared/reconciled against selectedResidueKeys when the drag
+    // ends. This is purely visual — the real 'sele' still comes from applyToggle.
+    @State private var dragSelKeys: Set<String> = []
 
     /// Flattened residue list (object order) for range/index mapping.
     private var flat: [SequenceResidue] { engine.sequences.flatMap { $0.residues } }
@@ -210,7 +216,9 @@ struct SequencePanel: View {
     }
 
     private func realResidueCell(_ r: SequenceResidue) -> some View {
-        let selected = r.isSelectable && engine.selectedResidueKeys.contains(r.selKey)
+        let selected = r.isSelectable
+            && (engine.selectedResidueKeys.contains(r.selKey)
+                || dragSelKeys.contains(r.selKey))
         return Text(r.oneLetter)
             .font(.system(size: 11, design: .monospaced))
             .foregroundColor(selected ? selectionFG : r.color)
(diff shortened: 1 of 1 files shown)


javierbq:
On-screen test (dev build RayMol-431, macOS host): **works** — dragging across residues in the sequence viewer now fills the selection background live during the drag, not just on release. #429 verified.

(Unrelated observation while testing with RT on: the Move gizmo leaves a lingering RT shadow on the object — filed separately as #433. Not caused by this PR.)

PR #435, opened 2026-09-09
fix(metal): derive key-light/shadow direction from cSetting_light (#434)

### LINKED ISSUE #434: macOS: Metal key light / shadow direction is hardcoded and ignores the 'light' setting (shadows barely visible)
### Summary
The native macOS Metal renderer uses a **hardcoded key-light direction** `normalize(0.4, 0.4, 1.0)` (eye space) for shading AND for shadow casting, and ignores PyMOL's `light` setting. Because that direction is near-frontal (dominant z), cast shadows fall behind the geometry and are barely visible ("no shadows"), and there is no way to swing the light to a raking angle to reveal them. Reported with `metal_raytrace` on, but it affects the raster shadow path too.

### Root cause (code)
The vector `normalize(0.4, 0.4, 1.0)` is hardcoded in six places in `layerGraphics/metal/RendererMetal.mm`:
- Lit shading (`L1`, the reflect/key light): lines ~4299, ~5794, ~6252
- Raster SSAO/shadow face gate (`Ldir`): line ~1280
- RT shadow/AO shader face gate (`Ldir`): line ~2229
- RT shadow ray direction (CPU `le`, transformed to model space as `lightDirModel`): line ~2660

Nothing feeds `cSetting_light` into the Metal renderer. The classic ray tracer uses it (`layer1/Ray.cpp`), and the convention is: **direction toward the light = `-normalize(cSetting_light)`** — which at PyMOL's default `light` equals the current hardcoded `(0.4,0.4,1.0)`, so respecting the setting is backward-compatible at default.

### Fix
Plumb `cSetting_light` into the Metal key-light direction so shading and shadows both follow it (and become user-adjustable via `set light, ...`). Read it in `layer1/SceneRender.cpp` next to the existing `setLightingParams` call and pass it to the renderer; replace all six hardcoded sites with the uniform. Keep CPU/GPU uniform layouts in sync (append scalar floats to avoid alignment surprises). Default behavior unchanged (default `light` reproduces the current look); users can now set a raking light to get visible, correctly-directed shadows.

Separate from #425/#427 (acceleration-structure bugs, fixed) — this is the light direction itself. Pre-existing on master.

Environment: macOS, host GPU with hardware RT, RayMol 1.10.0.

javierbq:
Fixes #434

## Problem
The native macOS Metal renderer hardcoded the key-light direction `normalize(0.4, 0.4, 1.0)` (eye space) for **both** shading and shadow casting, ignoring PyMOL's `light` setting. Because that direction is near-frontal (dominant z), cast shadows fell behind the geometry and were barely visible, and there was no way to swing the light to a raking angle. This affected both the raster shadow path and the `metal_raytrace` path.

## Fix
Plumb `cSetting_light` into the Metal renderer. `SceneRenderMetal` reads it next to the existing `setLightingParams` call and passes it through a new `Renderer::setKeyLightDir()` virtual. `RendererMetal` stores `_keyLightEye = -normalize(light)` — the direction **toward** the light, matching the classic ray tracer's convention (`layer1/Ray.cpp`). PyMOL's default `light` (`[-0.4,-0.4,-1.0]`) reproduces the historical `normalize(0.4,0.4,1.0)`, so **default output is unchanged**; `set light, ...` now repositions shading and shadows together.

## Sites changed
`layerGraphics/metal/RendererMetal.mm` (unless noted):
- **Lit shading `L1`** — `vbo_shade`, the sphere impostor fragment, the cylinder impostor fragment, and the second `LightU` shader (surface/tube).
- **Raster SSAO/shadow face gate** (`PostU`, was `Ldir` at ~1280).
- **RT shadow/AO face gate** (`RTU`, was `Ldir` at ~2229).
- **RT CPU `le`** — now built from `_keyLightEye`; the existing `lm = inv * le` still produces `lightDirModel` for the hardware-RT shadow rays.
- **`cap_fill` default `LightU` literal** — updated to the default key-light constants to keep the field count matching.
- **`layer1/SceneRender.cpp` `SceneBuildLightViewProjEye`** — the raster shadow-map **camera** direction was *also* hardcoded to `normalize(0.4,0.4,1.0)`. This is an extra site beyond the six the issue enumerated, but it is required: without it, `set light` would move the face gate and shading but the actual cast shadows (which come from the shadow map) would stay put. Now derived from `-normalize(cSetting_light)`, consistent with the shaders.

## Keeping CPU/GPU layouts in sync
The key-light is carried as **three appended scalar floats** (`klx, kly, klz`) rather than a `float3`/`float4`, to avoid 16-byte alignment surprises. Each MSL struct and its C++ mirror was edited in lockstep — identical field order, tail padding recomputed to the same 16-byte-multiple size on both sides:
- `LightU` (both MSL definitions) + the 5 CPU `_lt` initializers.
- `SphereU`: MSL + C++ mirror, pad `_pad0,_pad1` → `_pad0,_pad1,_pad2` (192 → 208 bytes).
- `CylU`: MSL + C++ mirror, pad → 3 floats (208 → 224 bytes).
- `PostU`: MSL + C++ mirror, `pad0` → `pad0,pad1` (144 → 160 bytes).
- `RTU`: MSL + C++ mirror, the 3 existing `pad` floats were repurposed as `klx,kly,klz` (size unchanged at 240 bytes).

## Verification
- **Verified:** both build stages are green — `bash swiftui/build_macos.sh` (C++ core, exit 0) then `xcodebuild ... -scheme PyMOLViewer_macOS -configuration Debug` → **BUILD SUCCEEDED**.
- **NOT verified here:** on-GPU rendering. These MSL shaders are compiled at runtime from source strings, so a CPU/GPU struct-layout mismatch would only surface live on the GPU, not at compile time. The layouts were kept byte-for-byte in sync by hand (see above), but the actual shadow/shading result on hardware still needs a live check (default look unchanged + `set light` swings the shadows). I have no hardware-RT GPU in this environment to confirm.

### DIFF
--- a/layer1/SceneRender.cpp
+++ b/layer1/SceneRender.cpp
@@ -1968,7 +1968,15 @@ static glm::mat4 SceneBuildLightViewProjEye(PyMOLGlobals* G, float* outRadius = 
   }
   if (radius < 1.0f)
     radius = 1.0f;
-  glm::vec3 Ldir = glm::normalize(glm::vec3(0.4f, 0.4f, 1.0f)); // toward light
+  // Direction TOWARD the key light in eye space = -normalize(cSetting_light).
+  // PyMOL's default `light` reproduces the historical normalize(0.4,0.4,1.0), so
+  // the shadow-map camera (and thus where cast shadows fall) now follows `light`
+  // and stays consistent with the shaders' key-light direction.
+  const float* lightv = SettingGetGlobal_3fv(G, cSetting_light);
+  glm::vec3 Ldir(-lightv[0], -lightv[1], -lightv[2]);
+  if (glm::length(Ldir) < 1e-6f)
+    Ldir = glm::vec3(0.4f, 0.4f, 1.0f); // degenerate light: historical fallback
+  Ldir = glm::normalize(Ldir);          // toward light
   glm::vec3 up = (Ldir.z * Ldir.z < 0.81f) ? glm::vec3(0, 1, 0)
                                            : glm::vec3(1, 0, 0);
   glm::vec3 lightPos = centerEye + Ldir * (radius * 2.0f);
@@ -2232,6 +2240,11 @@ void SceneRenderMetal(PyMOLGlobals* G)
         specReflect,
         specPower,
         SettingGetGlobal_f(G, cSetting_metal_sss_wrap));
+    // Key-light direction: feed cSetting_light so shading AND shadows follow it
+    // (and become user-adjustable via `set light`). The renderer stores
+    // -normalize(light) as the direction toward the light; PyMOL's default light
+    // reproduces the historical hard-coded normalize(0.4,0.4,1.0).
+    G->Renderer->setKeyLightDir(SettingGetGlobal_3fv(G, cSetting_light));
     // MSAA: 4x when metal_msaa is on, otherwise single-sample. The renderer
     // stashes this and applies it at the next beginLiveFrame (no encoder open),
     // so toggling at runtime never mismatches an in-flight encoder.
--- a/layerGraphics/Renderer.h
+++ b/layerGraphics/Renderer.h
@@ -397,6 +397,13 @@ public:
   {
   }
 
+  // PyMOL key-light direction (cSetting_light). The direction toward the light
+  // used for shading and shadow casting is -normalize(light). Default: no-op
+  // (the GL renderer reads cSetting_light itself). The Metal renderer stores
+  // it and feeds it into its lit/shadow/RT shaders so shading and shadows
+  // follow the light setting. lightv is a 3-vector (eye space, PyMOL's light).
+  virtual void setKeyLightDir(const float* /*lightv*/) {}
+
   // MSAA sample count for the scene (opaque) pass. SceneRenderMetal calls this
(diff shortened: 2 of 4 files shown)


javierbq:
Merged current master and **resolved the conflict** with #437 (both PRs appended fields to the RT composite `RTU` uniform). The MSL `RTU` struct had already merged cleanly; I mirrored that layout on the CPU side: the key-light fields `klx/kly/klz` take the three tail slots #437 had used for `pad0/pad1/pad2`, followed by #437's `aoCrease / aoCreaseRadiusPx / aoExemptEnabled / pad3` — 12 floats after `lightViewProj`, 16-byte aligned, identical field order CPU↔MSL. Also dropped the now-unused `u.pad0=pad1=pad2=0` assignment (those slots are the key-light now) and kept #437's `aoMaskReady`/crease population.

Verified: two-stage build **BUILD SUCCEEDED**; the merged RT path renders coherently (no uniform-layout corruption), and `light` still steers the shadows (default vs raking = ~45% of viewport pixels change). #434 behavior intact and #437's SSAO-crease-under-RT preserved.

PR should now be mergeable (was CONFLICTING).

Issue #420, opened 2026-09-07
Sets 6/6: folder and score-file import, ensemble ghost, iPad and iPhone layouts


javierbq:
Part of #421 (tracking). Spec §4.6, §4.7, §5 flow 2, §8 decision 5, wireframe frame 6. Depends on #418.

## Scope

- Import: drop a folder of `*.pdb` / `*.cif` with a `score.sc`, `*.csv` or `*.json` onto the window, or `set_import`; columns inferred from the file with a mapping sheet; structures referenced and stored lazily; FASTA and a3m import produce sequence-only sets.
- Ensemble ghost: a viewport toggle that draws the filtered set as thin CA traces at low alpha from a precomputed coordinate array, not objects; capped count with sampling above the cap; one focus entry solid on top.
- Thumbnails: lazily rendered from the peek object for rows scrolled past, capped cache, explicit "render all" action; stored in the container only when the user asks.
- iPad: the drawer is the existing bottom panel (`panelFrac`) with the same tabs; the inspector's SETS section in the side sheet.
- iPhone: sets open as a full-screen sheet with Table and Plot only; tap = peek, swipe actions = stage / star / reject.

## Test gate

Import parser tests for Rosetta score files and CSV with mixed types. Ensemble cap and sampling tests. iOS UI tests on the local scheme for the sheet.

## Done when

A ColabFold or Rosetta output folder becomes a browsable set in one drop, and "show me everything that passes" is honest and cheap on every size class.

Issue #419, opened 2026-09-07
Sets 5/6: Sequences tab replaces the strip, per-residue heat strips, Lineage tab


javierbq:
Part of #421 (tracking). Spec §4.4, §4.5, §8 decision 2, wireframe frames 3 and 4. Depends on #417.

## Scope

- Move `SequencePanel.swift` into the drawer as the Sequences tab. With no set open it shows exactly what the strip shows today (scene objects, alignment gaps, residue click selects atoms), so the behaviour in #380 is preserved.
- Rows are scene objects plus selected or filtered entries, aligned by shared parent position when they share a parent, gap-aware when an alignment object is enabled. Above a few hundred rows, collapse to a consensus/logo band with the selected rows on top.
- Per-residue metrics (pLDDT, MPNN native fit and certainty) as a heat strip under each row, using the domains `DesignColor` already defines, loaded lazily from the blob store.
- `LineageView`: left-to-right DAG target → backbones → sequences → folds, node size or fill from a chosen metric, hollow for rejected or failed; clicking a node filters the Table to its descendants.
- Remove the old sequence-strip slot and its `sequenceVisibleKey` handling in favour of the drawer tab.

## Test gate

Existing sequence-panel tests move with the view and stay green. New: alignment-by-parent layout, collapse threshold, lineage descendant filter.

## Done when

Spec §5 flow 3: stage three, compare their sequences with confidence strips under each, and read in Lineage which backbone each came from.

javierbq:
Implemented and merged: PR #455 (5/6 — Sequences tab, per-residue heat strips, Lineage tab), tracked as done in #421.

Closing as completed. This did not auto-close because the PR targeted `feat/sets` rather than the default branch, and GitHub only fires `Closes #N` on a default-branch merge — so every finished item in this epic has been sitting open. Shipping to `master` still depends on the `feat/sets` merge (#420 and #463 outstanding).

Issue #380, opened 2026-09-04
Sequence viewer lists disabled objects — should only show enabled ones


javierbq:
## Problem

The sequence viewer shows a row for **every** loaded molecular object, whether or not it is enabled. Uncheck an object in the object panel and it disappears from the viewport, but its sequence strip stays in the sequence panel. With a handful of structures loaded (or a session where you toggle between two models) the panel fills with sequences for things that are not on screen, and the residue rows you can actually click no longer correspond to what you are looking at.

Desktop PyMOL's `seq_view` only draws enabled objects; the panel should match.

## Root cause

`modules/pymol/appkit_sequence.py` `poll()` collects names without the enabled filter:

```python
names = list(cmd.get_names('public_objects') or [])
```

The module's own docstring already states the intended behavior — *"Emits one residue row per **enabled** molecular object"* — so the implementation has drifted from the documented contract. `cmd.get_names` takes `enabled_only`, which maps to the `rec->visible` check in `ExecutiveGetNames` (`layer3/Executive.cpp`), and PyMOL already cascades enable/disable down groups, so a member of a disabled group is correctly excluded too.

Nothing downstream compensates: `PyMOLEngine.fetchSequences()` just calls `appkit_sequence.poll()`, and `parseSequencePanelFeedback` in `swiftui/PyMOLViewer/Panels/SequencePanel.swift` renders whatever rows arrive with no visibility check.

## Expected

- Only enabled molecular objects get a sequence row.
- Disabling an object removes its row; re-enabling brings it back. `SequencePanel` already re-fetches on `engine.objects` change and `ObjectEntry` is `Equatable` over `isEnabled`, so the toggle should refresh without extra plumbing.
- Theme-studio preview (`preview=True`, reserved `__theme_preview`) is unaffected.
- Alignment gap layout keeps working for the enabled members of an enabled alignment.

## Repro

1. `fetch 1ubq, async=0` and `fetch 5hbh, async=0`
2. Open the sequence viewer — two rows, as expected.
3. Uncheck `5hbh` in the object panel. Viewport shows only 1ubq; the sequence panel still shows both.

javierbq:
Correction to the root-cause note above: `enabled_only=1` alone is **not** sufficient. I claimed PyMOL cascades enable/disable down a group so members of a disabled group drop out — it does not clear the member's own flag. Verified against a running build:

```
group g1, 5hbh
disable g1
[...]

Issue #417, opened 2026-09-07
Sets 3/6: inspector SETS section, Data drawer with the Table tab, peek/stage/pin, .raymol document type


javierbq:
Part of #421 (tracking). Spec §4.1, §4.2, §4.6, wireframe frames 1 and 5. Depends on #415; useful with #416.

## Scope

- `ObjectPanel.swift`: `SetEntry` (deliberately not an `ObjectEntry`, like `AlignmentEntry`) and a SETS section: name, count, ranking-column histogram, running badge while a batch lands, "open" affordance. Clicking a set row opens the drawer on that set. The set's group row shows an "n of N staged" badge.
- `Panels/DataDrawer.swift` with `SetTableView`: virtualized rows, leading star and stage-state columns (● ○ ◐), columns driven by `MetricSpec` (label, units, format, default sort direction, color ramp), footer with Stage / Pin / Export / Save view, "n of N" count. Keys: ↑↓ peek, space stage, `s` star, `x` reject.
- Peek on hover or arrow key, drawn superposed on the reference as a dashed/ghost style; one hidden object reused.
- `PyMOLEngine`: a read-only SQLite connection (`SetsStore`), refreshed on a `SETS:v<n>` change marker; no per-tick polling; `appkit_sets.py` emits only the marker and active-set/peek state.
- `PanelLayout`: `dataDrawerVisibleKey`, `dataDrawerFracKey`; the drawer takes the sequence strip's slot (the strip moves into it as a tab in #419; until then it stays where it is).
- `PyMOLApp.swift` and `project.yml`: `.raymol` UTType and document type, Save / Save As routing, the #349 replace-session guard extended to the new extension.
- With no set open, nothing changes: SETS collapses when empty, the drawer is hidden by default.

## Test gate

Swift: `SetTableModelTests` (sort, format, budget), `PanelLayoutTests` for the new keys, `OpenFilesTests` for `.raymol`. Functional check on the VM harness with a 1000-entry set: scrolling and peeking stay smooth, the Objects panel shows only staged objects.

## Done when

The triage loop works end to end for RFD3 output: sort, peek through the top twenty, star, stage, save, reopen.

javierbq:
Implemented and merged: PR #446 (3/6 — inspector SETS section, Data drawer with the Table tab, peek/stage/pin, .raymol document type), tracked as done in #421.

[...]

PR #456, opened 2026-09-16
Sets: the scene sequence strip is a drawer band, not a tab

javierbq:
## Summary

**The strip is an Object view; the Sequences tab is an Entry view. They may not be
alternatives.** #419 moved the scene rows into the drawer as the top half of the
Sequences tab and made `wantsSequences` mean "drawer up AND tab == Sequences" — so the
scene sequences became mutually exclusive with Table, Plot and Lineage. The scene rows
show the sequences of the **enabled scene objects**; the tab's rows show the **contents
of a set**. That is the Entry/Object split this whole epic is built on, and two
different nouns cannot be two states of one control: "what residues am I looking at" and
"which candidates pass" are the ordinary case of two questions asked at once.

Spec §8 decision 2's "two sequence views would compete" is satisfied by there being
exactly **one** scene view — and this PR puts it in the drawer's chrome, as a band above
the tab bar, drawn whatever tab is selected.

```
BEFORE (#419)                              AFTER (#456)
┌───────────────────────────────────┐      ┌───────────────────────────────────┐
│ DATA · rfd3_a1  [Table][Plot]     │      │ DATA · rfd3_a1            ⌘2  ✕   │
│                 [Sequences][Line] │      │ 4HHB_target  A  VLSPADKTNVKAAWG… │ ← band
│ ── the Sequences tab ──           │      │ d_0417       A  SPIGGQWTAGLRRIK… │   (any tab)
│ 4HHB_target  A  VLSPADKTNVKAAWG…  │      │ ───────────────────────────────── │
│ d_0417       A  SPIGGQWTAGLRRIK…  │      │ [Table][Plot][Sequences][Lineage] │
│ ── split ──                       │      │            filter: plddt>80 ⌕    │
│ d_0116  PWLRHWWEACGVSLGCPNFFKM…   │      │ ☆ 👁 id     plddt ▾ iptm  rmsd    │
│ d_0120  PKMTMNLVCCNYDVWFYKRLTA…   │      │ ★ ● d_0185  95.0   0.81  2.9      │
│                                   │      │   ⋮        [Stage] [Send to ▾]   │
│ …and the Table is NOT on screen   │      │                                   │
└───────────────────────────────────┘      └───────────────────────────────────┘
```

- **⌘2 toggles the band**, independently of the tab. So do the rail's **Seq** pill, the
  new **⌘2** switch in the drawer header, and `View ▸ Show/Hide Sequences`. None of them
  changes the tab any more.
- **The band is `SequencePanel()` unchanged** — the compatibility contract #419 was
  careful to preserve: #380 enabled-only, the ruler, real guide-atom colours, click and
  drag selection, both feedback parsers. Its height behaviour is #419's too: the strip's
  own `min(max(objects,1),5) * 30 + 30` ideal (now `PanelLayout.sequenceBandIdealHeight`),
  a 400 pt ceiling, resizable (the hairline under it is a `VSplitView` divider), and an
  `.id()` so it re-adopts the ideal height when objects load.
- **With no set open the drawer is just the band** — no tab bar, no tab content. That is
  today's strip in its new home, and it replaces the old "the Sequences tab works
  without a set" rule: `worksWithoutASet` left `DataDrawerTab` and is now
  `SequenceBandView.worksWithoutASet`.
- **The Sequences tab is only the set's entry rows** now, with their heat strips and the
  consensus band.
- **Turning the band off gives its height back** to the tab content.
- **The tab strip moved from the header row to the filter row**, where spec §4's sketch
  has it: the header is the drawer's, the tab row is the set's. Net chrome cost with the
  band off is zero.
- **iOS is untouched.** The strip is still its own pane there until #420, `sequenceVisible`
  is still its flag, and both migrations stay `#if os(macOS)`.

## Migration

There are two migrations now and a user can arrive with either combination, so
`PanelLayout.sequenceBandMigration` is a pure function with one test per case
(`SequenceBandMigrationTests`), the way `sequenceStripMigration` already is. It takes
#419's answer for the same launch, which is what tells the arrival routes apart.

| # | #419 flag | `sequenceVisible` | drawer | tab | → band | why |
|---|---|---|---|---|---|---|
| 1 | not yet run | `true` | — | — | **on** | they came straight from a build with a strip and it was up; a visible pane must not vanish because it moved |
| 2 | not yet run | `false` | — | — | off | "off" is a choice too |
| 3 | not yet run | absent | — | — | off | fresh install, or never touched the strip |
| 4 | spent | — | open | Sequences | **on** | #419 put them there; the scene rows are what they are looking at right now, and those rows are leaving that tab |
| 5 | spent | — | closed | Sequences | off | a closed drawer is not showing anything, whatever tab it remembers |
| 6 | spent | — | open | Table/Plot/Lineage | off | they are triaging; ⌘2 is how they ask for sequences |
| 7 | #456 already run | — | — | — | stored value, nothing written | or turning the band off would be undone next launch |

Two further tests run it against a real `UserDefaults` domain in launch order (#419
first, because it writes `dataDrawerVisible`; #456 reading what it left): the
never-migrated user with the strip up ends with both the drawer and the band on, and the
already-#419-migrated user on the Sequences tab is carried forward to the band.

Persisted under its own key, `raymol.panels.sequenceBandVisible`, with
`raymol.panels.sequenceBandMigrated` beside it; both are in `PanelLayout.allKeys`.

## The drawer height ceiling with the band on

The band takes height from the **drawer**, not from the viewport column, so
`drawerColumnUsed` is unchanged — what moves is the drawer's own minimum. The review
caught the constants not meaning their names, so they are now measured against the views
and everything else is derived from them:

```
macDrawerTabRowHeight   24   DataDrawer.tabRow (tabs + filter)
macDrawerTableChrome    62   SetTableView: column header 20 + histograms 16 + footer 24 + 2 hairlines
macDrawerRowHeight      22   one table row
macDrawerHeaderHeight   26   "DATA · <set>   ⌘2 ✕", above the band

macMinTabContentHeight  = 24 + 1 + 62 + 22          = 109   (was 70 — below the chrome alone)
macDrawerChromeHeight   = 26 + 1 + 24 + 1 + 62      = 114   (was 73)
macMinDrawerHeight      = 114 + 22                  = 136   (was 96, pre-existing from #417)
minDrawerHeight(band)   = 136 + 40 + 1              = 177
```

`PanelLayout.drawerTableRows` states the whole thing in the number the review found the
bug in, and `testADrawerThatFitsAlwaysDrawsAtLeastOneTableRow` asserts it across
600–1200pt, console open and closed, one to five objects: **any drawer the layout agrees
to draw must draw a row of the table under the band.**

`defaultDrawerHeight(bandVisible:)` adds the band's allowance (a two-object scene, 90pt)
to the untouched 220pt default, so turning the sequences on out of the box does not cost
the user their table — a two-object scene keeps exactly the rows it had with the band
off. A drawer the user has **sized** still keeps its size and the band takes from the tab
content; that is what "turning it off gives the height back" means, and the pair of
screenshots below shows it (band on → 2 rows, band off → 5, same drawer).

**At 800pt** with the console at its 130pt default and the rail up, the column leaves a
234pt ceiling — above the 177pt minimum, so `drawerFits` passes and the untouched drawer
is 311pt: a two-object band at its full 90pt ideal with four table rows under it, a
five-object band clamped to 98 with one row. The band yields by exactly what the tab half
needs, never more. Below roughly 720–740pt with the console open the ceiling falls into
[136, 177) — the drawer fits, the drawer-plus-band does not — and that is the hint's
range, which is why it now offers **Hide Sequences** next to Hide Console. The drag
divider clamps to the same `minDrawerHeight`, so the drag and the hint cannot disagree.

## Test output

**Swift** — `bash swiftui/build_macos.sh` then

```
xcodebuild -project swiftui/PyMOLViewer.xcodeproj -scheme UnitTests_macOS \
  -destination 'platform=macOS,arch=arm64' -configuration Debug \
  -derivedDataPath swiftui/build_mac_dd -skipPackagePluginValidation \
  -skipMacroValidation test
```

```
Test Suite 'All tests' passed
   Executed 932 tests, with 4 tests skipped and 0 failures (0 unexpected)
** TEST SUCCEEDED **
```

910 → **932**, +22, exactly the 22 test functions this PR adds. The `SequencePanel` /
`SequenceRowModel` tests that covered the scene rows inside the tab were *adapted*, not
weakened: `testTheSceneRowsKeepTheStripsHeightFormula` asserts the same five numbers
(60/90/180/180/60) on `SequenceBandView.sceneHeight` and additionally that the view's
formula and `PanelLayout`'s are the same one.

**Python** — under a private shadow tree (`~/.cache/raymol-shadow-456`).

Owned files, one process:

```
pymol -ckqy testing/testing.py --run testing/tests/sets \
  testing/tests/test_appkit_sets.py testing/tests/test_appkit_sequence.py \
  testing/tests/test_appkit_sequence_panel.py
→ Ran 202 tests ... OK   (unittest half)
→ 97 collected, 97 passed  (pytest half)
→ grep -cE 'FAILED \[|^(FAIL|ERROR):' = 0
```

The full CI list (all 39 entries from `.github/workflows/raymol-embedded-tests.yml`), one
process:

```
→ Ran 1163 tests ... FAILED (errors=9, skipped=2)   (unittest half)
→ 360 collected, 360 passed                        (pytest half)
→ grep -cE 'FAILED \[|^(FAIL|ERROR):' = 9   → all 9 are the documented local baseline
```

The 9 are pre-existing and are the numpy-ABI mismatch between Homebrew's `_cmd` and the
fork under this harness — **6 in `raymol/shadow_extent.py`** (`testCacheTracksContentChanges`,
`testCameraMotionDoesNotMoveTheBox`, `testDisabledObjectsAreExcluded`,
`testEmptySceneHasNoExtent`, `testSolventIsExcluded`,
`testSolventOnlySceneFallsBackToAllAtoms`) and **3 in `generate/`**
(`testTheTargetIsUnmovedInTheEmittedObject`, `testAMovedTargetIsReadWHEREITISNotWhereItWasStored`,
`testTheANIMATIONRefusesAnObjectThisRunDidNotSeed`). No Python file is touched by this PR
(`git diff --stat` is 9 Swift files and 2 docs), so none of them can be its doing.

## Screenshots

Re-taken after the review, **at the app's default drawer height with the console open**
— which is what a user gets after the migration, and is why the blocker was invisible
the first time (those shots had the console closed and the drawer hand-dragged to
~354pt). Second instance `RayMol-456.app` on `RAYMOL_MCP_PORT=48999`, driven over MCP:
4HHB as the target, a generated 1024-entry `rfd3_a1` set with pLDDT/ipTM/RMSD/len and
per-residue pLDDT arrays, two entries staged, filter `plddt > 80` → **212 of 1024**.
1400×900 unless noted. Local paths (not uploaded to GitHub):

- `/tmp/raymol456-table-band-on.png` — **the point of the change**, and the blocker's
  fix: three scene rows in the band AND the candidate table with rows and its full
  footer, at the drawer's default height. Before the fix this drawer showed the column
  header and nothing else.
- `/tmp/raymol456-band-off.png` — the same drawer with ⌘2 off: five rows instead of two.
  The band costs three rows and gives them straight back.
- `/tmp/raymol456-sequences-band-on.png` — the Sequences tab, band above, entry rows
  with their pLDDT heat strips below (filter tightened to `plddt > 93` for this one so
  the rows draw rather than collapsing to the 200-row consensus band).
- `/tmp/raymol456-no-set.png` — the drawer with no set open: header, band, nothing else.
  No tab bar. This is today's strip, and it looks like it.
- `/tmp/raymol456-1332x771-band-on.png` — **the app's own persisted window size**, the
  review's worst case. Band plus three table rows and the footer.
- `/tmp/raymol456-no-room-hint-1332x771.png` — the same window with the console up and
  dragged tall: the column leaves 140pt, which fits the drawer (136) but not the drawer
  plus the band (177), so the hint fires — and offers **Hide Sequences** beside Hide
  Console. Clicking Hide Console gives the shot above. This is the band never being able
  to make the drawer it lives in unreachable, in practice.

## Deferred

- **iPad / iPhone (#420).** The drawer is still macOS-only, so the band is too. The strip
  remains its own pane on iOS with `sequenceVisible` as its flag, and both migrations are
  `#if os(macOS)` — an ungated run would spend a migration that day's layout still needs.
- The band's height is not separately persisted; the `VSplitView` position is per-session
  and the ideal height is re-adopted when the object count changes, which is #419's
  behaviour unchanged.
- The Sequences tab still keeps its own horizontal scroller, independent of the band's.
  The two column spaces are different things (a scene row is numbered by the structure's
  own `resi`; an entry row is padded against its parent group), so a shared scroll would
  imply a correspondence that is not there.
- While the "needs more room" hint is showing, `wantsSequences` is still true, so the
  500 ms sequence poll runs for a band that is not drawn. Harmless (it is the same cost
  the band-visible case pays) and `wantsSequences` has no way to see the window height;
  left alone rather than threaded through.
- `macDrawerTableChrome` and friends are constants measured against the views, not read
  from them. A view that changes its own header height would need this updated with it;
  `testTheDrawersMinimumsAreItsActualPartsAddedUp` at least makes the sum explicit.


javierbq:
Review addressed — pushed as `a158063c4`. Blocker 1 and recommendations 2 and 3 are in, plus four of the nits.

**Blocker 1.** You were right and the diagnosis was exact. `macMinTabContentHeight` is now `macDrawerTabRowHeight 24 + hairline + macDrawerTableChrome 62 + macDrawerRowHeight 22` = **109**, the constants are measured against the views rather than guessed, and `macDrawerChromeHeight` / `macMinDrawerHeight` are derived from the same parts (114 / **136** — the #417 undercount fixed too, since this PR propagated it by subtraction). `minDrawerHeight(bandVisible: true)` = 177.

On the either/or: I took **both**, because they answer different cases. `defaultDrawerHeight(bandVisible:)` adds a two-object band's allowance to the untouched 220pt default, so an out-of-the-box band is free — a two-object scene keeps exactly the rows it had with the band off. A drawer the user has *sized* keeps its size and the band takes from the tab content, which is what "turning it off gives the height back" has to keep meaning. The honest division is still what happens once the ceiling bites.

`PanelLayout.drawerTableRows(drawerHeight:bandObjects:)` states the regression in your number, and `testADrawerThatFitsAlwaysDrawsAtLeastOneTableRow` asserts it over 600–1200pt × console open/closed × 1–5 objects.

**Recommendation 2** was the better catch of the two, and it is a #419 bug this PR would have shipped on top of: `migrateSequenceStrip` now writes `dataDrawerTabKey` when it chooses Sequences. `testAMigratedUserWhoNeverTouchedATabStillGetsTheBand` omits the key, as #419 really leaves it.

**Recommendation 3.** `DataDrawer.height` is required; `restoredSequenceBandVisible` lost the parameter it ignored.

**Nits taken:** consensus-band symbols renamed (`consensusBandSetID`, `loadSequencesForConsensusBand`, `ConsensusBandTests`) so "band" means the scene band everywhere else; `DataDrawer.content(hasSet:bandVisible:)` is a pure decision that *reads* `SequenceBandView.worksWithoutASet`, so the property is load-bearing and its test is no longer `true == true`; ⌘2 no longer closes a drawer ⌘4 opened (`drawerClosesWithBand`, tested); the filter bar outranks the tab strip for width; ObjectPanel's stale comment; the §4 sketch gains the staged/budget readout.

**Nits left, with reasons** (both now in Deferred): the 500 ms poll still runs while the hint is showing — `wantsSequences` has no way to see the window height and the cost is the same one the band-visible case pays; and the measured constants are constants, so a view that changes its own header height needs them updated with it (the new test at least makes the sum explicit rather than the difference).

**Mutation check** — each fix has a test that fails without it:

| reverted | failing tests |
|---|---|
| `macMinTabContentHeight` → 70 | 23 assertions |
| drop the `dataDrawerTabKey` write | `testAMigratedUserWhoNeverTouchedATabStillGetsTheBand` |
| `defaultDrawerHeight` ignores the band | `testTheUntouchedDrawerDrawsRowsWithTheBandOn` |
| `drawerClosesWithBand` → `!hasSet` | `testHidingTheBandOnlyClosesTheDrawerItOpened` |

**Swift gate:** 932 tests, 4 skipped, 0 failures, `** TEST SUCCEEDED **` (910 → 932, +22 = the 22 test functions this PR adds). **Python:** owned files 202 unittest OK + 97 pytest passed, loose grep 0; full CI list 1163 unittest + 360 pytest, loose grep 9 — the documented baseline (6 `shadow_extent`, 3 `generate`, numpy ABI).

**Screenshots re-taken at the default drawer height with the console open**, which is the state your point about invisibility was about. Two of them are the fix and the guard rail:

- `/tmp/raymol456-table-band-on.png` — band with three scene rows AND the table with rows and its full footer, at the default height. Same drawer with ⌘2 off (`/tmp/raymol456-band-off.png`) is five rows: the band costs three and gives them straight back.
- `/tmp/raymol456-1332x771-band-on.png` — the app's own persisted size: band plus three rows and the footer.
- `/tmp/raymol456-no-room-hint-1332x771.png` — same window, console up and dragged tall: the column leaves 140pt, which fits the drawer (136) but not drawer-plus-band (177), so the hint fires and offers **Hide Sequences**. That is the [136, 177) gap being reachable rather than a dead end.
- `/tmp/raymol456-sequences-band-on.png`, `/tmp/raymol456-no-set.png` as before.

Still local paths only — `gh` has no image-upload endpoint.

PR #469, opened 2026-09-18
Objects panel: Copy to Object + Extract on selection rows

### LINKED ISSUE #461: Objects panel: add PyMOL's "copy to object ▸ new" to the A menu
Reported by Gabriel Rocklin (Slack, 2026-09-16):

> It would be great if you could imiplement pymol's "copy to object->new"

## What's missing

Legacy PyMOL's selection/object action menu has **copy to object ▸** — a submenu listing every existing object plus **new**, wired to `cmd.copy_to()`:

- `modules/pymol/menu.py:1182` (and :1208, :1269, :1776) — `[ 1, 'copy to object', lambda: copy_to(self_cmd, sele) ]`, built by `copy_to()` at `modules/pymol/menu.py:1136`
- `modules/pymol/editing.py:3285` — `copy_to(name, selection, rename='chain segi ID', zoom=-1, quiet=1)`

The command exists in the fork and works from the console. What's absent is the UI entry: the SwiftUI action ("A") menu in `swiftui/PyMOLViewer/Panels/ObjectPanel.swift` has no copy/extract family at all. `baseActionMenuItems` (line 684) goes Zoom/Orient/… ▸ Preset ▸ Find ▸ Align ▸ Generate ▸ Hydrogens ▸ State ▸ …, and selections only add **Remove Atoms** (`actionMenuItems(isSelection:)`, line 835).

Note the semantic difference from the neighbouring legacy entries, which are also missing: `copy_to` **copies** atoms into a target object (renaming chain/segi/ID to avoid collisions) and leaves the source intact, whereas `extract` **moves** them out.

## Proposed

Add to the selection row's A menu (and, where it makes sense, to object rows):

- **Copy to Object ▸** — dynamic submenu of currently-loaded molecule objects, plus a separator and **new…**. `ActionMenuItem` already has a dynamic-submenu case for exactly this shape (the group submenu added for #255, line 678).
- **new…** should prompt for a name, defaulting the way `copy_to(None, …)` does (`obj01`, `obj02`, …).
- Likely worth adding **Extract Object** alongside it for parity, but `copy to object ▸ new` is the requested one.

## Acceptance

- Selecting atoms, then A ▸ Copy to Object ▸ new creates a new object containing a copy of the selection; the source object is unchanged.
- Copying into an existing object merges the atoms in with chain/segi/ID renamed so nothing collides.
- The new object appears in the Objects panel immediately and is enabled.

### LINKED ISSUE #468: Objects panel: add PyMOL's "extract object" to the selection A menu
Reported by Gabriel Rocklin (Slack, follow-up to the request in #461):

> also it would be great to get 'extract sele to new object' back

## What's missing

Legacy PyMOL's selection action menu has **extract object**, wired to `cmd.extract()`:

- `modules/pymol/menu.py:1183` (and :1209, :1777, :1836) — `[ 1, 'extract object', 'cmd.extract(None,"'+sele+'",zoom=0)' ]`
- `modules/pymol/menu.py:47` — `extract(self_cmd, sele)` builds the richer **Extract ▸** submenu used elsewhere: `object`, `extend 1`, `byres extend 1`, each a `cmd.create(None, <sel>, extract=<sele>, zoom=0)`
- `modules/pymol/creating.py:1050` — `cmd.extract(name, selection, ...)`

The command exists in the fork and works from the console (`extract new_obj, sele`). What's absent is the UI entry: the SwiftUI action ("A") menu in `swiftui/PyMOLViewer/Panels/ObjectPanel.swift` has no copy/extract family at all. `baseActionMenuItems` (line ~684) runs Zoom/Orient/… ▸ Preset ▸ Find ▸ Align ▸ Generate ▸ Hydrogens ▸ State ▸ …, and selection rows only add **Remove Atoms** (`actionMenuItems(isSelection:)`, line ~725).

## Relationship to #461

Sibling, not duplicate — the two differ in whether the source object keeps its atoms:

| | source object | command |
|---|---|---|
| **copy to object ▸ new** (#461) | atoms stay | `cmd.copy_to()` |
| **extract object** (this issue) | atoms are **moved out** | `cmd.extract()` |

#461 mentions extract as an aside; this issue makes it the deliverable, since it is what was explicitly asked for. They should probably land together as one "Copy / Extract" section of the A menu.

## Proposed

On selection rows in the Objects panel A menu, add an **Extract ▸** submenu next to the #461 copy entries:

- **object** → `create None, <sele>, extract=<sele>, zoom=0`
- **extend 1** → `create None, (<sele>) extend 1, extract=<sele>, zoom=0`
- **byres extend 1** → `create None, byres ((<sele>) extend 1), extract=<sele>, zoom=0`

Naming follows PyMOL's `None` default (`obj01`, `obj02`, …). If a name prompt is wanted, reuse the rename modal path (`engine.pendingRename`) rather than inventing a second one.

Object rows should not get this — extract is only meaningful for a sub-selection of an existing object, matching desktop PyMOL.

## Acceptance

- With atoms selected, A ▸ Extract ▸ object creates a new object holding those atoms, and they are **gone** from the source object.
- `extend 1` / `byres extend 1` pull in the bonded neighbourhood as upstream does.
- The new object appears in the Objects panel immediately, enabled, without a manual refresh; the camera does not jump (`zoom=0`).
- The now-empty selection is handled sanely (dropped rather than left pointing at nothing).

javierbq:
Implements both halves of the request Gabriel made in Slack: PyMOL's `copy to object ▸ new` (#461) and, in his follow-up, `extract sele to new object` (#468). They ship together because they are one section of the action menu and only make sense read against each other.

| menu entry | source object | command |
|---|---|---|
| **Copy to Object ▸ New Object…** | keeps its atoms | `create <name>, (<sele>), zoom=0` |
| **Copy to Object ▸ \<existing\>** | keeps its atoms | `copy_to <obj>, (<sele>), zoom=0, quiet=0` |
| **Extract ▸ object** | atoms **move out** | `cmd.extract(None, "(<sele>)", zoom=0)` |
| **Extract ▸ extend 1** | atoms **move out** | `cmd.create(None, "((<sele>) extend 1)", extract="<sele>", zoom=0)` |
| **Extract ▸ byres extend 1** | atoms **move out** | `cmd.create(None, "(byres ((<sele>) extend 1))", extract="<sele>", zoom=0)` |

Both are gated to **selection rows**, matching desktop PyMOL, where they sit in the `else:` branch of an `if object:` — `modules/pymol/menu.py:1183` (sele_action), `:1209`, `:1777`, `:1836`. The section goes in the object-management tail, above Move to Group, which is where upstream keeps it too.

## Decisions worth a look

- **Extract auto-names, Copy asks.** Extract is one click and lets PyMOL mint `obj01`/`obj02` exactly as the desktop item does — it's the fast path Gabriel asked for. The copy prompts (prefilled with the same `objNN` default, computed client-side to mirror `ExecutiveGetUnusedName`), per #461's acceptance: a copy is usually kept and compared against its source, and `obj01` tells you nothing at that point.
- **Extract deletes the emptied selection.** `cmd.extract` only cleans up its own temporary, so the user's named selection would otherwise survive with every atom gone — a 0-atom row that still looks live. Same thing "Remove Atoms" already does.
- **The widened variants print a deprecation warning.** They must spell out `extract="<sele>"` rather than the boolean `extract=1`, because the atoms to remove are a subset of the atoms to copy — that asymmetry is the whole point of "extend 1". `cmd.create` calls that form deprecated and says so on the console. Upstream's own menu has the same wart; the alternative is not having the feature.
- **Copy-to targets are not filtered by parent object.** Upstream hides the objects a selection already lives in (`menu.py:1136`). The panel's row model doesn't record which objects a selection spans, so that filter is left out rather than guessed at.
- `ActionMenuItem` / `actionMenuItems` become internal so menu composition is testable, and `ObjectPanel.body` is split in two — a third `.alert` in one chain puts it past the Swift type-checker's budget, the same reason other views in this file are broken up.

## Verification

- **`xcodebuild test -scheme UnitTests_macOS`: 725 tests, 0 failures, 4 skipped.** 20 of those are new (`ObjectPanelCopyExtractTests`): menu composition (selection-only, group beats selection, all three Extract scopes, ordering, Remove Atoms preserved), every command string, and the `objNN` namer including gap-filling and the past-99 case.
- **Semantics checked against a real PyMOL**, driving the exact emitted strings through `cmd.do()` on 1ubq — 16/16: extract moves atoms out and the source loses exactly them; the widened variants copy more than they remove; `copy_to` leaves the source untouched and the target gains the atoms; a disjunctive selection (`resi 1-5 or resi 20-25`) survives the parenthesising intact.

Not verified: the menu was **not** exercised in the app. The `mac-vm-pool` MCP server is refusing connections in this environment, so the `mac-vm-test` path CLAUDE.md calls for was unavailable. The iOS slice also could not be link-checked — `build_ios/libpymol_core.a` in this checkout predates the Metal renderer (no `beginLiveFrame` symbol), a stale local artifact unrelated to this change; the Swift compiles, only the C++ link fails. **Worth a hands-on click-through of both submenus before merge.**

Closes #461
Closes #468


javierbq:
## Verified in the app

The menu is now driven under a synthetic finger on an iPhone 17 Pro simulator — `ObjectActionMenuUITests`, 4/4 passing:

```
testSelectionMenuOffersCopyAndExtract   passed (15.958 s)
testObjectMenuOffersNeither             passed (17.906 s)
testExtractCreatesObjectAndDropsSelection passed (25.852 s)
testCopyToObjectListsTargetsAndPrompts  passed (31.953 s)
```

Run with:

```
cd swiftui && xcodebuild test -scheme UITests_iOS -sdk iphonesimulator \
  -destination 'platform=iOS Simulator,name=iPhone 17 Pro' \
  -only-testing:PyMOLViewerUITests/ObjectActionMenuUITests
```

Each test attaches labelled screenshots (`1-selection-menu`, `3-extract-submenu`, `5-copy-to-submenu`, `6-new-object-prompt`, `4-after-extract`, `7-after-copy`) — open the `.xcresult`, or export them with `xcrun xcresulttool export attachments`.

What they show: **Extract ▸** opens with `object` / `extend 1` / `byres extend 1`; picking `object` leaves `obj01` under OBJECTS and **"No selections"** below it, with the console echoing `cmd.extract(None, "(mysele)", zoom=0); cmd.delete("mysele")`. **Copy to Object ▸** opens with `New Object…` above a separator and the loaded object `mol`; `New Object…` prompts *Copy "mysele" to a new object / The selected atoms stay in their current object too*, prefilled `obj01`; confirming runs `create obj01, (mysele), zoom=0` → "found 78 atoms", and `mol`, `obj01` and `mysele` all remain.

## One thing worth a decision

On iPhone the section is **below the fold of a scrolling menu** — past Preset, Find, Align, Generate, Hydrogens, State, Sequence, Movement, Masking, Compute. The UI test only found it after scrolling. That placement matches upstream, but upstream's selection menu is a fraction of this one's length, so "where PyMOL keeps it" doesn't buy the same discoverability here. If this is meant to be a daily-use verb, it probably wants to move up — say next to Remove Atoms, or just under the Zoom/Orient/Center block. Happy to move it; it's a one-line change to where `copyExtractMenuItems` is spliced in.

## Fixes that came along

- `build_ios/libpymol_core.a` was two months stale and predated the Metal renderer, which is why the iOS slice wouldn't link. Rebuilt via `swiftui/build_ios.sh` (its CMake cache also pointed at a deleted worktree — cleared).
- Added a `UITests_iOS` scheme. The generated `PyMOLViewer_iOS` scheme carries the macOS-only `PyMOLViewerTests`, so `xcodebuild` refuses to build it for `iphonesimulator` ("Could not find test host"), and `-skip-testing:` doesn't help because the failure is in host resolution before any test runs. **The UI tests in this repo were not runnable as documented before this.**
- `ActionMenuButton` gains `accessibilityIdentifier("actionMenu.<name>")`, matching the existing `colorMenu.*` hook.

javierbq:
Worked through the Copilot review. Three of the four findings were real — verified each against a real PyMOL rather than taking the described mechanism on trust, which mattered, because the most serious one was right about the danger and wrong about the cause. Pushed in d9d26b5.

### ✅ Name validation (Critical) — real, but not for the stated reason

The claim was that an existing name makes `create` *add the atoms to that object*. It doesn't:

```
before: atoms 660 states 1
PyMOL> create prot, (mysele), zoom=0
after:  atoms 660 states 1   names ['prot']
```

Nothing happens. No merge, no new object, no error, no console line — **Copy just silently fails**, which is worse than a merge because there's nothing to notice or undo. The comma case is real and separately ugly:

```
PyMOL> create foo, bar, (mysele), zoom=0
ValueError: invalid literal for int() with base 10: '(mysele)'
```

And an illegal character doesn't error at all — the engine rewrites it (`my obj!` → `my_obj`), so the panel ends up showing a name nobody typed.

Fixed by checking against PyMOL's own legal set (`ObjectMakeValidName`: `A–Z a–z 0–9 + - . ^ _`) and against the existing namespace — objects, selections and groups share it. The check runs in the action body, not only as `.disabled()` on the button, since the action body is what builds the command string. The alert's subtitle now says which rule was broken instead of just greying out.

### ✅ copy_to hides the source (Moderate) — real, and worse than described

Confirmed at `editing.py:3309`. Reproduced:

```
PASS  BUG reproduced: plain copy_to disables the source   enabled=[]
PASS  fix: source stays VISIBLE                           enabled=['prot', 'other']
PASS  fix: target still gained the atoms
PASS  fix: source keeps its atoms
PASS  fix: an already-hidden source stays hidden
```

Note the first line: `[]` — it disabled *everything*, not just the source. On the desktop this is a momentary thing inside a menu; here the source's row stays in the panel with its checkbox clearing on its own, right after a menu item that promises a copy. It reads as the copy having eaten the original. Now the enabled sources are captured before the copy and switched back after; already-hidden ones stay hidden.

### ✅ Pending targets (Moderate) — real

`isPending` exists on `ObjectEntry` and those rows are `.disabled()` + 35 % opacity everywhere else in the panel. They're placeholders for a running inference job with no atoms yet. Dropped from the target list.

### ❌ Doubled separators (Nit, 2 votes) — not reproducible

`baseActionMenuItems` already ends `… Compute, .separator, .moveToGroup, .separator, Rename…`. Splicing `copyExtractMenuItems + [.separator]` **at** the `moveToGroup` index puts the new rows *between* the existing separator and the new one:

```
Compute │ ─── │ Copy to Object, Extract │ ─── │ Move to Group │ ─── │ Rename
```

One divider each. Dropping the trailing separator as suggested would instead leave Extract butted directly against Move to Group. The [Extract submenu screenshot](https://github.com/javierbq/RayMol/pull/469#issuecomment-5731675516) shows the single divider under Compute. Added `testNoDoubledSeparators`, which walks all three menu shapes asserting no two separators are adjacent — so if this ever does regress, it fails in CI rather than in review.

### Also fixed

- The UI test's run instructions named `PyMOLViewer_iOS` — the scheme that can't build for the simulator, which is why `UITests_iOS` exists in this PR. Corrected, with a note saying why.
- `menuItem()` returned a non-existent element when scrolling ran out, so the caller's `.tap()` failed with "Failed to tap Any (First Match)" — naming neither the row nor the real cause. It now `XCTFail`s by name.

### Verification

- `UnitTests_macOS`: **732 tests, 0 failures**, 4 skipped (was 725 — 7 new: name legality, collisions, the default name always being usable, copy-visibility ordering, separators).
- `UITests_iOS/ObjectActionMenuUITests`: **4/4 passing** against the rewritten copy command.
- Command semantics re-checked in a real PyMOL, 5/5 on the visibility fix as shown above.

PR #473, opened 2026-09-19
DOF: aperture 0 means no blur, not maximum blur (#472)

### LINKED ISSUE #472: macOS: metal_dof_aperture 0 gives MAXIMUM blur (0 is the renderer's "unset" sentinel)
### Summary
`metal_dof_aperture = 0` produces the **strongest** blur the slider can make, not the weakest. The renderer treats `0` as an "unset" sentinel and substitutes the default of `14.0`, so dragging the **DOF aperture (blur)** slider down to 0 jumps from near-sharp (1 px bokeh radius) to maximum bokeh. The slider therefore reads as monotonic from 1→40 but wraps around at its own minimum.

### Steps to reproduce
1. Load any structure and enable **Depth of field** (`metal_dof`).
2. `set metal_dof_aperture, 1` — the scene is essentially sharp (1 px blur radius).
3. `set metal_dof_aperture, 0` — the scene becomes heavily blurred, identical to `metal_dof_aperture, 14`.
4. Same via the Scene panel: drag **DOF aperture (blur)** to its minimum.

### Expected
Aperture 0 means zero blur (a closed aperture) — visually equivalent to turning DOF off. At minimum it should be the *least* blurred point of the slider, never the most.

### Cause
The sentinel test in the DOF post pass, `layerGraphics/metal/RendererMetal.mm:2802`:

```objc
u.maxRadiusPx = ((_dofAperture > 0.0f) ? _dofAperture : 14.0f) * pixelRadiusScale();
```

`0` fails the `> 0.0f` test and falls through to the hard-coded default `14.0f`. The same expression is duplicated in the export-alpha matte pass (`RendererMetal.mm:3039`), so hi-res Copy/Save output shows it too.

`metal_dof_range` has the identical trap at `RendererMetal.mm:2798` / `:3038` (`(_dofRange > 0.01f) ? _dofRange : 14.0f`). It is currently unreachable from the UI only because that slider is clamped to `min: 1`, but it is reachable from the command line: `set metal_dof_range, 0`.

The aperture slider is the one that exposes it, since it allows 0 — `swiftui/PyMOLViewer/Panels/ObjectPanel.swift:318`:

```swift
SceneParam(setting: "metal_dof_aperture", label: "DOF aperture (blur)", kind: .slider, min: 0, max: 40, ...)
```

This is already a known hazard on the animation side: `modules/pymol/raymol_scene_anim.py:42` carries a `_FLOOR` of `0.02` for both settings, commented *"Values at/below the renderer's sentinel are reinterpreted as 14 (MAXIMUM blur) … so an interpolated fade must never reach them."* Without that floor, a scene transition fading DOF out would snap to full blur on its last frame. The floor is a workaround for this bug, not a fix.

### Proposed fix
Move the sentinel out of the valid value range in all four places — treat only a *negative* value as "unset":

```objc
u.maxRadiusPx = ((_dofAperture >= 0.0f) ? _dofAperture : 14.0f) * pixelRadiusScale();
u.focusRange  = (_dofRange   >= 0.0f) ? _dofRange   : 14.0f;
```

so `0` means an actual zero radius / zero range. The gather shader already handles a sub-pixel `maxRadiusPx` (all taps land in the center texel), so aperture 0 degrades to a pass-through; `focusRange` still needs a small epsilon guard against division by zero in the CoC term. Follow-ups once the sentinel moves:

- drop or relax `_FLOOR` in `raymol_scene_anim.py` so DOF fades can reach a true zero;
- consider lowering the `metal_dof_range` slider `min` to 0 for consistency.

Note `metal_dof_focus = 0` is a *deliberate* and documented sentinel ("0 = auto-focus on the center of interest", resolved in `SceneRender.cpp:2061-2072`) and should stay as is — this issue is only about aperture and range, where 0 is a meaningful physical value.

### Environment
macOS, Metal renderer, `master`.

javierbq:
Fixes #472.

### The bug
`metal_dof_aperture = 0` produced the **strongest** blur the slider can make, not the weakest: the Metal DOF pass tested `_dofAperture > 0.0f` and fell through to the hard-coded `14.0f` default, so the **DOF aperture (blur)** slider wrapped around at its own minimum — 1 is near-sharp, 0 is full bokeh.

`metal_dof_range` carried the identical test (`> 0.01f`) and was only unreachable from the UI because that slider was clamped to `min: 1`; `set metal_dof_range, 0` hit it from the command line.

### The fix
Move the "unset" sentinel out of the valid value range — only a **negative** value now resolves to the default:

```objc
u.focusRange  = (_dofRange    >= 0.0f) ? _dofRange    : 14.0f;
u.maxRadiusPx = ((_dofAperture >= 0.0f) ? _dofAperture : 14.0f) * pixelRadiusScale();
```

in both the DOF post pass and the export-alpha matte pass. Aperture 0 is a closed aperture, range 0 an instant falloff — both meaningful values a user can dial in.

At aperture 0 the DOF pass is now **skipped entirely** rather than run as an expensive no-op. (It was already a no-op numerically: every CoC evaluates to 0, `dof_composite` mixes back to the sharp sample and the near-field coverage stays 0 — it just paid for a 16–96-tap gather per pixel first.) The export-alpha matte needed no change; its halo gather was already guarded on `maxRadiusPx > 0.5`.

### Fallout
`raymol_scene_anim._FLOOR` (0.02 for aperture and range) existed *only* to keep an interpolated DOF fade from landing on the sentinel and flashing full blur on its last frame. With zero no longer poisoned it is gone: a fade dissolves to a true `_OFF_APERTURE = 0.0`, and `value_at` is a plain lerp again. The `metal_dof_range` slider drops to `min: 0` for consistency, and both sliders' help text says what 0 does.

Not touched: `metal_dof_focus = 0`, which is a deliberate, documented "auto-focus on the center of interest" sentinel resolved per frame in `SceneRender.cpp`.

### Testing
- `testing/tests/test_raymol_scene_anim.py` — 52 passed. The three tests that pinned the old floor now pin the new contract: `value_at` reaches exactly 0, and the DOF fade ramps from a genuinely closed aperture.
- Live app: Debug build of this branch (`RayMol-472.app`), 1ubq with cartoon + sticks, `metal_dof` on, `metal_dof_range 6`, `metal_dof_quality 4`, autofocus off, focus auto. Screenshots of the Metal view at three apertures:
  - **14** — heavy bokeh, the whole molecule dissolved.
  - **1** — nearly sharp, a faint halo on the out-of-focus sticks.
  - **0** — fully sharp, indistinguishable from DOF off (the pass is skipped).

  Monotonic from 0 to 14, which is the whole point: before this change, 0 rendered identically to 14.


javierbq:
Verified on a Debug build of this branch (`RayMol-472.app`, 1ubq + cartoon/sticks, `metal_dof` on, range 6, quality 4, focus auto):

| `metal_dof_aperture` | live Metal view |
|---|---|
| 14 | heavy bokeh, molecule fully dissolved |
| 1 | near-sharp, faint halo on the defocused sticks |
| 0 | completely sharp — identical to DOF off, DOF pass skipped |

Monotonic across the slider's whole range now. Before the change, 0 and 14 rendered the same image.

The screenshots are host captures of the live Metal view (the MCP `capture_viewport` path is the CPU ray tracer and does not run the DOF post pass, so it cannot show this); they are in the session, not attachable from the CLI.

